PR #14, opened 2018-06-18
Installer script

francoiscampbell:
Vary basic, only works by cloning the repo because it's not open source. Also clobbers anything already at `/usr/local/bin/yvm` and `~/.yvm`

## QA
- run `make install`
- open a new shell tab, you should see 
```
running yvm script
This shell is interactive, declaring yvm function only
```
- `type yvm` should say `yvm is a shell function from /usr/local/bin/yvm`
- `yvm`, `/usr/local/bin/yvm`, and `make yvm-test` should launch yvm


brandonbaksh:
QA looks good

PR #22, opened 2018-06-19
Yvm exec

brandonbaksh:
- implemented yvm exec command
- usage: `yvm exec <version> <yarn commands>`

## Dev prep
- use yvm install to get 1.7.0 and 1.0.2

## QA Steps
- `make install` and open a new terminal (or `source /usr/local/bin/yvm`)
- use `yvm exec 1.7.0 versions`
- use `yvm exec 1.0.2 versions`
- ensure the correct output from both
- navigate to fe-thm, use `yvm exec 1.0.2 upgrade-interactive` and ensure it works as expected

### DIFF
--- a/src/commands/exec.js
+++ b/src/commands/exec.js
@@ -1,0 +1,25 @@
+const fs = require('fs')
+const os = require('os')
+const path = require('path')
+
+const getYarnCLI = version => (
+    path.resolve(getYarnPath(version), 'bin/yarn.js')
+)
+
+const getYarnPath = version => (
+    path.resolve(os.homedir(), `.yvm/versions/v${version}`)
+)
+
+const execCommand = (version, extraArgs) => {
+    const yarnPath = getYarnPath(version)
+    if(!fs.existsSync(yarnPath)) {
+        console.log(`Yarn v${version} not found. Please install it using: yvm install ${version}`)
+        return
+    }
+
+    const yarnCLI = getYarnCLI(version)
+    process.argv = ['',''].concat(extraArgs) //first two arguments are filler args [node version, yarn version]
+    require(yarnCLI)
+}
+
+module.exports = execCommand
--- a/src/commands/exec.js
+++ b/src/commands/exec.js
@@ -1,19 +1,21 @@
+/* eslint-disable global-require,prettier/prettier,import/no-dynamic-require */
 const fs = require('fs')
 const os = require('os')
 const path = require('path')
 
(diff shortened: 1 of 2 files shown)


src/commands/exec.js
```diff
@@ -0,0 +1,25 @@
+const fs = require('fs')
+const os = require('os')
+const path = require('path')
+
+const getYarnCLI = version => (
```
francoiscampbell: this isn't the CLI file any more, i'd just make a function called `runYarn` that massages the `argv` and calls `require(path.resolve(getYarnPath(version), 'bin/yarn.js'))`

src/index.js
```diff
@@ -16,9 +16,12 @@ module.exports = function(args) {
...
 
     argParser
-        .command('manage <version>')
-        .action(version => {
-            console.log(`Managing yarn v${version}`)
+        .command('exec <version>')
+        .action((version) => {
+            console.log(`Executing yarn command with version ${version}`)
+            const exec = require('./commands/exec')
+            const extraArgs = args.slice(4)  //returns args without [npm, start, exec, <version>, <command>]
```
francoiscampbell: seems `commander` can glob the rest if you use `...`: https://www.npmjs.com/package/commander#variadic-arguments

you could probably just do 
```
.command('exec <version> [extrasArgs...]')`
.action((version, extraArgs) => {
```

src/index.js
```diff
@@ -16,9 +16,12 @@ module.exports = function(args) {
         })
 
     argParser
-        .command('manage <version>')
-        .action(version => {
-            console.log(`Managing yarn v${version}`)
+        .command('exec <version>')
+        .action((version) => {
```
umar-tophat: I think you can add an extra arg on this line and it will contain the rest of the args to the command.

umar-tophat: something like `(version, rest) => {`

umar-tophat:
## ✅ QA

- `npm start exec <version> versions` shows the correct version for each version number
- If I use a version that has not been installed, I get a nice suggestion on how to install
- `upgrade-interactive` works as expected.

src/commands/exec.js
```diff
@@ -1,19 +1,21 @@
+/* eslint-disable global-require,prettier/prettier,import/no-dynamic-require */
```
francoiscampbell: why disable prettier?

brandonbaksh: rip i just copy pasta'd that from a different file

PR #29, opened 2018-06-19
Yvm remove

brandonbaksh:
## Dev prep
- npm install
- make install

## QA Steps
- install a version of yarn using `yvm install <version>`
- remove the version using `yvm remove <version>` and ensure that `~/.yvm/versions/v<version>` does not exist anymore

### DIFF
--- a/src/commands/remove.js
+++ b/src/commands/remove.js
@@ -1,0 +1,27 @@
+const fs = require('fs-extra')
+const os = require('os')
+const path = require('path')
+
+const yvmPath = path.resolve(os.homedir(), '.yvm')
+const versionRootPath = path.resolve(yvmPath, 'versions')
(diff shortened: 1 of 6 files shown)


src/commands/remove.js
```diff
@@ -0,0 +1,27 @@
+const fs = require('fs-extra')
+const os = require('os')
+const path = require('path')
+
+const yvmPath = path.resolve(os.homedir(), '.yvm')
+const versionRootPath = path.resolve(yvmPath, 'versions')
```
umar-tophat: This seems to be duplicated from install.js. Can we move it to a common file?

umar-tophat:
## ✅ QA 
`yvm remove` deletes version folder

PR #36, opened 2018-06-19
Yvm list

brandonbaksh:
## Description
- added `yvm list` command

### DevQA Prep
- `make install`

### DevQA Steps
- `yvm install 1.7.0`
- run `yvm list` and ensure it outputs that version
- run `yvm remove 1.7.0` and then `yvm list` and ensure it outputs that you don't have any versions of yarn

### DIFF
--- a/src/commands/list.js
+++ b/src/commands/list.js
@@ -3,7 +3,7 @@ const fs = require('fs')
 const { printVersions, versionRootPath } = require('../common/utils')
 
 const getYarnVersions = () => {
-    const re = /^v(\d+\.)(\d+\.)(\d+)/
+    const re = /^v(\d+\.)(\d+\.)(\d+)$/
     const validItems = []
     fs.readdirSync(versionRootPath).forEach(file => {
         if (re.test(file)) {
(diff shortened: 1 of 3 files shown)


umar-tophat:
## ✅ QA
- `yvm list` shows the installed versions of yarn

src/commands/list.js
```diff
@@ -0,0 +1,28 @@
+const fs = require('fs')
+
+const { printVersions, versionRootPath } = require('../common/utils')
+
+const getYarnVersions = () => {
+    const re = /^v(\d+\.)(\d+\.)(\d+)$/
```
umar-tophat: I think @francoiscampbell added this to utils: https://github.com/tophatmonocle/yvm/blob/master/src/util/version.js#L3

src/commands/list.js
```diff
@@ -0,0 +1,28 @@
...
+    const validItems = []
+
+    if (fs.existsSync(versionRootPath)) {
+        fs.readdirSync(versionRootPath).forEach(file => {
```
umar-tophat: You can use filter here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

src/commands/list.js
umar-tophat: You may need to merge master.

PR #39, opened 2018-06-19
Auto-install and use stderr

francoiscampbell:
## Description
- Auto install for ease
- Use stderr to avoid polluting `yarn` output

### DevQA Steps
- `make install`
- `yvm exec 1.7.0 versions` should install 1.7.0, then output the result of the command
- Running it again should use the installed version
- `yvm exec bin 2> /dev/null` should not print anything that `yarn` itself doesn't print

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,7 +5,6 @@
     "prettier"
   ],
   "rules": {
-    "no-console": 0,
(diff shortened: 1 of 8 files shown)


brandonbaksh:
LGTM ✅

PR #43, opened 2018-06-19
Use makefiles, Add test runner + codecov

jakebolam:


Makefile
```diff
@@ -1,3 +1,43 @@
+# so we can use `eslint` without ./node_modules/.bin/eslint
+SHELL := /bin/bash
+export PATH := $(shell npm bin):$(PATH)
+
+ARTIFACT_DIR?=artifacts
+TEST_REPORTS_DIR?=$(ARTIFACT_DIR)/reports
+BUILD_DIR?=$(ARTIFACT_DIR)/build
+
+IGNORE = test.js,js.snap,.md,__tests__,mockdata,__mocks__
+BABEL_FLAGS = src --out-dir $(BUILD_DIR) --copy-files --ignore $(IGNORE)
```
jakebolam: Setup this up for babel, will do the babel stuff in another PR

francoiscampbell: FYI, since we'll be using webpack, we won't need to manually call babel

jakebolam: Didn't know we were going down this path, can we remove this on that PR?

package.json
```diff
@@ -10,6 +10,17 @@
   "scripts": {
     "precommit": "lint-staged"
   },
+  "jest": {
+    "testPathIgnorePatterns": [
+      "/lib/",
```
francoiscampbell: we don't have `/lib/` do we?

Makefile
francoiscampbell: yup, i'll take care of it

Issue #51, opened 2018-06-20
Add short descriptions for each command


umar-tophat:
`.description('Some description')`

Issue #50, opened 2018-06-20
Show help string when yvm called without params


umar-tophat:
http://tj.github.io/commander.js/#Command.prototype.help

Issue #49, opened 2018-06-20
Add aliases for commands


umar-tophat:
- `i` from `install`
- `ls` for `list`
- `rm` for `remove`

PR #52, opened 2018-06-20
Improve command line UX

umar-tophat:
## Description

Addresses #49, #51, and #50 

- Add aliases for install, remove, and list
- Add descriptions for commands
- Show help when no params provided
- Update API docs

### DevQA Prep

- Please `make install`

### DevQA Steps

- Run `yvm`
- Verify help is shown, and that aliases and descriptions added
- Run `yvm i` and ensure it calls install
- Run `yvm rm` and ensure it calls remove
- Run `yvm ls` and ensure it calls list

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -27,6 +27,7 @@ module.exports = function dispatch(args) {
     /* eslint-disable global-require,prettier/prettier */
     argParser
         .command('install [version]')
+        .alias('i')
         .action(withRcFileVersion(version => {
             log(`Installing yarn v${version}`)
             const install = require('./commands/install')
@@ -35,6 +36,7 @@ module.exports = function dispatch(args) {
 
     argParser
         .command('remove <version>')
+        .alias('rm')
         .action(version => {
             log(`Removing yarn v${version}`)
             const remove = require('./commands/remove')
@@ -59,6 +61,7 @@ module.exports = function dispatch(args) {
 
     argParser
         .command('list')
+        .alias('ls')
         .action(() => {
             log(`Checking for installed yarn versions...`)
             const listVersions = require('./commands/list')
--- a/src/index.js
+++ b/src/index.js
@@ -76,6 +76,11 @@ module.exports = function dispatch(args) {
             listVersions()
         })
 
+    const noParams = !process.argv.slice(2).length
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -52,13 +60,16 @@ module.exports = function dispatch(args) {
     argParser
         .command('list-remote')
         .alias('ls-remote')
+        .description('List Yarn versions available to install.')
```
brandonbaksh: non blocking: this could be misleading since we do not remove versions they have already installed. Maybe just "List available Yarn versions" would be better? What do you think?

src/index.js
```diff
@@ -25,8 +25,13 @@ const withRcFileVersion = action => (maybeVersionArg, ...rest) => {
 
 module.exports = function dispatch(args) {
     /* eslint-disable global-require,prettier/prettier */
+    argParser
+        .description('Yarn Version Manager')
```
brandonbaksh: Could possibly add functionality to make `yvm` direct to `yvm --help`

brandonbaksh: Could possibly add functionality to make yvm direct to yvm --help

src/index.js
brandonbaksh: non blocking: this could be misleading since we do not remove versions they have already installed. Maybe just "List available Yarn versions" would be better? What do you think?

umar-tophat: I think "List available Yarn versions" might be too close in meaning to `yvm list`. But I agree that we need to think carefully about the terminology for the commands. Even the word "install" seems somewhat misleading to me.

src/index.js
brandonbaksh: this might require shell support - if so, we can do it in a separate pr

src/index.js
brandonbaksh: "List available Yarn version releases" ??

src/index.js
umar-tophat: I just pushed a commit that accomplishes this, but we don't have `use` listed yet. That could be addressed in separate PR

src/index.js
umar-tophat: Going to merge this for now, but will keep this in mind once YVM is in the wild.

PR #65, opened 2018-06-21
Add help command

umar-tophat:
## Description

Addresses #53 

- Add help command (which does the same thing as `yvm`, `yvm -h`, and `yvm --help`)
- Update API docs

### DevQA Prep

- Run `make install`

### DevQA Steps

1. Run `yvm help`
   - Verify that help text is outputted

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -32,7 +32,7 @@ Display the current version of Yarn.
 
 Uninstall the specified version of Yarn.
 
+### `help`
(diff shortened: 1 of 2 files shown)


brandonbaksh:
Do you know if there is a way to move `help`  out of the "commands:" section and into the "options" section? 

screen shot 2018-06-21 at 11 16 44 am

Issue #64, opened 2018-06-21
Make it easier to test changes to JS files


umar-tophat:
Before, we could run `node yvm.js` to test the JS code, but since we added Babel, we need to compile every time.

So far, I have just been running `make install` every time, but it would be nice to have either:

- `make watch` - to watch, rebuild and copy to `~/.yvm`
- symlink webpack build artifacts to `~/.yvm`

PR #66, opened 2018-06-21
easier development, make development better

jakebolam:
Fixes: #64 
bitmoji

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,9 @@ install: build
 .PHONY: install-watch
 install-watch: node_modules
 	mkdir -p ${CURRENT_DIR}/artifacts/webpack_build/
+	rm -rf ~/.yvm
 	ln -i -s ${CURRENT_DIR}/artifacts/webpack_build/ ${HOME}/.yvm
+	chmod +x ${HOME}/.yvm/yvm.sh
 	webpack --config webpack/webpack.config.base.js --watch
 
 .PHONY: yvm-test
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -42,6 +44,14 @@ help:
 install: build
 	@use_local=true scripts/install.sh
 
+.PHONY: install-watch
+install-watch: node_modules
+	mkdir -p ${CURRENT_DIR}/artifacts/webpack_build/
+	rm -rf ~/.yvm
+	ln -i -s ${CURRENT_DIR}/artifacts/webpack_build/ ${HOME}/.yvm
```
jakebolam: If we're deleting `.yvm` then there is no need for `-i` on this line, as that is prompting to do that

Makefile
```diff
@@ -26,10 +27,11 @@ help:
 	@echo "--------------------- Useful Commands for Development ----------------------"
 	@echo "make help                            - show tasks you can run"
 	@echo "make install                         - runs a set of scripts to ensure your environment is ready"
+	@echo "make install-watch                   - runs install, and watches code for local development"
 	@echo "make yvm-test                        - runs the yvm node app"
 	@echo ""
 	@echo "----------------------- Other Commands  -------------------------"
-	@echo "make build                            - runs eslint"
+	@echo "make build                           - runs eslint"
```
francoiscampbell: runs eslint?

PR #68, opened 2018-06-21
Dogfooding

francoiscampbell:
## Description

- Use yvm in our own Makefile

### DevQA Steps
- Remove all trace of `yvm` on your system
- Run the public install script: `curl https://raw.githubusercontent.com/tophatmonocle/yvm/master/scripts/install.sh | bash`
- `make clean`
- `make build` should use yarn 1.7.0 instead of npm


jakebolam:
YES

PR #71, opened 2018-06-21
Better interactive shell detection

francoiscampbell:
According to https://www.gnu.org/software/bash/manual/html_node/Is-this-Shell-Interactive_003f.html, the old and new methods should be equivalent, but on CircleCI, $PS1 is set even in non-interactive shells


brandonbaksh:
LGTM ✅

PR #74, opened 2018-06-22
Source yvm in .bash_profile during install

umar-tophat:
## Description

- Add line to source yvm in `.bash_profile

### DevQA Prep

- Open bash
- Start fresh:
	- `rm /usr/local/bin/yvm`
	- `rm -rf ~/.yvm`
	- Remove yvm versions from PATH
	- Remove lines that source yvm in `~/.bash_profile` if it exists

### DevQA Steps

- Run `make install`
- Run `type yvm`
	- Verify that it says `yvm is a function`
- Check that yvm is sourced in `~/.bash_profile`
- Check that you can use `yvm use` correctly

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -55,4 +55,9 @@ if ! grep -q "${executable_source_string}" ~/.zshrc; then
     echo ${executable_source_string} >> ~/.zshrc
 fi
 
+if ! grep -q "${executable_source_string}" ~/.bash_profile; then
+    [ -z "${added_newline}" ] && echo '' >> ~/.bash_profile
+    echo ${executable_source_string} >> ~/.bash_profile
+fi
+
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -55,4 +55,9 @@ if ! grep -q "${executable_source_string}" ~/.zshrc; then
     echo ${executable_source_string} >> ~/.zshrc
 fi
 
+if ! grep -q "${executable_source_string}" ~/.bash_profile; then
```
brandonbaksh: Lines 46 to 51 might need to go here for ~/.bash_profile?

brandonbaksh: or at the very least you might need to reset `added_newline` ?

umar-tophat: Whoops. I prematurely merged this. I can address your comment in a follow up PR

PR #75, opened 2018-06-22
Validate install version is >= 1.0.0

brandonbaksh:
## Description
- Moved request for yarn tags to be a util function
- Updated list-remote to use the util function
- Updated install to use the util function to check if the user's version matches what we support
- support for versions greater than 1.0.0

### DevQA Steps
- `make install`
- `yvm install 1.7.0` should work
- `yvm install 1.9.0` should not work
- `yvm install `0.28.4` should not work
- ensure `yvm list-remote` works as expected (displays all valid versions over 1.0.0)

### DIFF
--- a/src/commands/install.js
+++ b/src/commands/install.js
@@ -7,6 +7,7 @@ const log = require('../common/log')
 const {
     versionRootPath,
     getExtractionPath,
+    getVersionsFromTags,
     yvmPath,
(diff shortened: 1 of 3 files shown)


umar-tophat:
## ✅ QA

- `yvm list-remote` only shows versions >= 1.0.0
- `yvm install` only allows installing valid versions

PR #78, opened 2018-06-22
Add `yvm use` to help and warn if user tries to run directly using Node

umar-tophat:
## Description

Addresses #60. 

### DevQA Prep

- Run `make install`

### DevQA Steps

- Run `yvm help`
	- Verify that `use` command is listed with message
- Run `node ~/.yvm/yvm.js use`
	- Verify that user is warned that they shouldn't call the JS file directly

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -88,15 +88,6 @@ argParser
     })
 
 argParser
-    .command('list-remote')
-    .alias('ls-remote')
-    .description('List Yarn versions available to install.')
-    .action(() => {
-        const listRemote = require('./commands/listRemote')
-        listRemote()
-    })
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -73,15 +78,6 @@ argParser
...
 
-argParser
-    .command('list-remote')
-    .alias('ls-remote')
-    .description('List Yarn versions available to install.')
-    .action(() => {
-        const listRemote = require('./commands/listRemote')
-        listRemote()
-    })
-
```
umar-tophat: This was duplicated in master. So I am just cleaning it up

Issue #72, opened 2018-06-21
Consistent formatting for list commands 


umar-tophat:
`yvm list` and `yvm list-remote` list version numbers in two different ways: one with "v" prefix and one without. We should make the listing consistent with the commands that the user may run next (eg. `yvm use 1.7.0` or `yvm exec 1.7.0`), so we should remove the "v" prefix.

You can use `stripVersionPrefix` in `src/common/utils.js`

umar-tophat:
![screen shot 2018-06-21 at 3 40 25 pm](https://user-images.githubusercontent.com/38886386/41741706-dca4ac72-7569-11e8-9d7a-a4db6d24a354.png)

PR #79, opened 2018-06-22
Remove 'v' prefix from versions in `yvm list`

umar-tophat:
## Description

Addresses #72 

### DevQA Prep

- Please `make install`

### DevQA Steps

- Run `yvm list` and then `yvm list-remote`
	- Verify that version formatting matches


brandonbaksh:
LGTM

PR #80, opened 2018-06-22
prevent removing currently active version

nepodmitljivi:
### DevQA Prep

- Please `make install`

### DevQA Steps
- Install and use one version of yarn
- Try running `yvm remove <version>`, it should fail
- Install and use a different version
- Try deleting previous yarn version, it should succeed


umar-tophat:
✅ QA LGTM

PR #81, opened 2018-06-22
Prevent yvm use when not a shell script; add auto-install to yvm use

francoiscampbell:
### DevQA Prep

- Please `make install`

### DevQA Steps
- Delete all your yarn versions
- `unset -f yvm`
- `yvm use` should fail
- `source /usr/local/bin/yvm`
- `yvm use` should auto-install 1.7.0


umar-tophat:
✅ QA looks good!

Issue #77, opened 2018-06-22
`make install-watch` should reset link `/usr/local/bin/yvm` when it is stopped


umar-tophat:

jakebolam:
Fixed in #85

PR #85, opened 2018-06-22
Update `make install-watch` and `install.sh` to clean up symlinks

umar-tophat:
## Description

- Rather than link the entire `artifacts/webpack_build` folder to `~/.yvm`, I decided to link the files that are actually changing (ie. `yvm.js` and `yvm.sh`). 
- Reasoning behind this that subsequent install commands will install into `artifacts/webpack_build` and removing the link will remove the versions you installed.

### DevQA Steps

- Run `make install-watch`
	- Verify that changes to code are updated without having to re-run install
- Stop watcher, run `make install`
	- Verify that links are removed, and normal install is completed

### DIFF
--- a/scripts/install-watch.sh
+++ b/scripts/install-watch.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+CURRENT_DIR=$(pwd)
+
+# Need to delete files, otherwise links are not created
+rm ${HOME}/.yvm/yvm.sh
+rm ${HOME}/.yvm/yvm.js
+
+# Link
+ln -s ${CURRENT_DIR}/artifacts/webpack_build/yvm.sh ${HOME}/.yvm/yvm.sh
+ln -s ${CURRENT_DIR}/artifacts/webpack_build/yvm.js ${HOME}/.yvm/yvm.js
+chmod +x ${HOME}/.yvm/yvm.sh
+
+# Start Webpack in watch mode
+webpack --progress --config webpack/webpack.config.dev.js --watch
--- a/scripts/install-watch.sh
+++ b/scripts/install-watch.sh
@@ -2,14 +2,48 @@
 
 CURRENT_DIR=$(pwd)
 
-# Need to delete files, otherwise links are not created
-rm ${HOME}/.yvm/yvm.sh
-rm ${HOME}/.yvm/yvm.js
+# Check if already watching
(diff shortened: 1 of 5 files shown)


scripts/install-watch.sh
```diff
@@ -0,0 +1,15 @@
...
+rm ${HOME}/.yvm/yvm.sh
+rm ${HOME}/.yvm/yvm.js
+
+# Link
+ln -s ${CURRENT_DIR}/artifacts/webpack_build/yvm.sh ${HOME}/.yvm/yvm.sh
+ln -s ${CURRENT_DIR}/artifacts/webpack_build/yvm.js ${HOME}/.yvm/yvm.js
+chmod +x ${HOME}/.yvm/yvm.sh
+
+# Start Webpack in watch mode
+webpack --progress --config webpack/webpack.config.dev.js --watch
```
francoiscampbell: how is the previous yvm recovered? maybe we should backup the current files and restore them after the `webpack` returns?

umar-tophat: Issue is how to execute command after Webpack. If you send Interrupt signal, the lines of the script that come after are not executed.

francoiscampbell: you can trap the interrupt and clean up: https://stackoverflow.com/questions/22752461/continuing-a-script-after-ctrl-c

scripts/install-watch.sh
```diff
@@ -0,0 +1,49 @@
...
+# Link
+ln -s ${CURRENT_DIR}/artifacts/webpack_build/yvm.sh "${HOME}/.yvm/yvm.sh"
+ln -s ${CURRENT_DIR}/artifacts/webpack_build/yvm.js "${HOME}/.yvm/yvm.js"
+chmod +x "${HOME}/.yvm/yvm.sh"
+
+# Trap interrupt signal
+int_trap() {
+    echo "Stopped Webpack"
+}
+trap int_trap INT
```
jakebolam: woah

scripts/install-watch.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Start Webpack in watch mode
+webpack --progress --config webpack/webpack.config.dev.js --watch
```
jakebolam: This works just like this? Sweet

jakebolam:
Fixes #77

PR #91, opened 2018-06-25
Shortcut for latest version

francoiscampbell:
# WIP

### DevQA Prep


.yvmrc
```diff
@@ -1 +1 @@
-1.7.0
+1.6.0
```
francoiscampbell: this is just a test

francoiscampbell:
closed because it's not super necessary and made arg parsing annoying. could be revisited later

PR #92, opened 2018-06-25
Add update-self command

francoiscampbell:
# WIP

### DevQA Prep


src/yvm.sh
```diff
@@ -90,6 +90,8 @@ yvm_() {
             exit 1
         fi
         yvm_use $2
+    elif [ "${command}" = "update-self" ]; then
```
jakebolam: Can we add docs on it? To `yvm help`

scripts/install.sh
```diff
@@ -62,3 +62,4 @@ if ! grep -q "${executable_source_string}" ~/.bash_profile; then
 fi
 
 echo "yvm successfully installed in ${YVM_DIR} with the 'yvm' command aliased as ${executable_alias_path}"
+echo 'Open another terminal window to start using it, or type `source ~/.yvm/yvm.sh`
```
francoiscampbell: should use ${YVM_DIR}

docs/faq.md
```diff
@@ -8,3 +8,7 @@ description: Common questions and answers
 
 ## How does yvm work?
 
+## I get the message `Do not call yvm.js directly!`
+This means you're running the yvm.js script directly and not the shell function.
+Try opening another terminal window, or type `source ~/.yvm/yvm.sh`
```
francoiscampbell: should use ${YVM_DIR}

Issue #108, opened 2018-07-03
Running `yvm exec` in a nested directory, says it can't find the .yvmrc file


jakebolam:
**Describe the bug**
Running `yvm exec` in a nested directory fails saying it can't find the .yvmrc file

**To Reproduce**
1. Create a valid .yvmrc file
2. Create a new directory, enter that sub-directory
3. Run `yvm exec list`
3. See error

**Expected behavior**
yvm should search up the tree for .yvmrc files

**Additional notes**
This has introduced other behaviour (such as having a root level `yvm.config.js` file, or declaring in `package.json`. Do we want this?

PR #109, opened 2018-07-03
Add support for search up directories for .yvmrc (as well as looking in package.json etc)

jakebolam:
## Description
Fixes: Running `yvm exec` in a nested directory, says it can't find the .yvmrc file #108 

## DevQA
Execute `yvm exec --version` in a directory with a `.yvmrc`, in a directory outside, and in a sub directory. Verify behaviour is correct.

### DIFF
--- a/src/util/version.js
+++ b/src/util/version.js
@@ -1,19 +1,18 @@
-const fs = require('fs')
+const cosmiconfig = require('cosmiconfig')
+const log = require('../common/log')
 
 function isValidVersionString(version) {
     return /\d+\.\d+\.\d+/.test(version)
 }
 
 function getRcFileVersion() {
-    try {
-        const rcFileContents = fs.readFileSync('.yvmrc', 'utf-8')
-        return rcFileContents.replace('\n', '')
-    } catch (e) {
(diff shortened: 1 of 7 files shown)


jakebolam:
[fix yarn linking in fe-libraries](https://trello.com/c/a0DVYl3l/168-fix-yarn-linking-in-fe-libraries)

src/util/version.js
```diff
@@ -1,19 +1,18 @@
...
 function getRcFileVersion() {
-    try {
-        const rcFileContents = fs.readFileSync('.yvmrc', 'utf-8')
-        return rcFileContents.replace('\n', '')
-    } catch (e) {
-        if (e.code === 'ENOENT') {
-            return null
-        }
-        throw e
+    const explorer = cosmiconfig('yvm')
```
dat2: We should add docs to readme showing that `.yvmrc`, `.yvmrc.js`, `.yvmrc.yaml` are all accepted now 👍

dat2:
QA LGTM

The `yvm use` and `yvm which` commands are done within the shell script though, and they don't have access to automatic yvmrc searching 😞

PR #118, opened 2018-07-09
Move PATH editing logic to JS

### LINKED ISSUE #112: yvm use and yvm exec - behaviour for getting config differ
**Describe the bug**
yvm use, looks in the current directory for a `.yvmrc` file. yvm exec does a more indepth tree and file search

**Expected behavior**
They should behave the same

francoiscampbell:
## Description

- Changes `yvm use` to use `yvm get-new-path` under the hood, moving most of the logic into JS for consistency
- Closes #112

### DevQA Steps
- `make install-watch`
- Open a new shell
- `yvm use` should still work
- `yvm get-new-path 1.7.0` should return a PATH string for yarn 1.7.0
- `yvm get-new-path` should use the version in .yvmrc
- `yvm use 1.6.0` then `yvm get-new-path 1.7.0` should edit the existing PATH string, not prepend a new segment to it

### DIFF
--- a/src/commands/getPath.js
+++ b/src/commands/getPath.js
@@ -1,0 +1,28 @@
+const path = require('path')
+
+const { versionRootPath, yvmPath } = require('../common/utils')
+
+const getPath = (version, rootPath = yvmPath) => {
+    const splitPath = process.env.PATH.split(path.delimiter)
+    const destPath = versionRootPath(rootPath)
+
+    const newPathSegment = `${destPath}/v${version}/bin`
+    let hadExistingPath = false
+
+    for (let i = 0; i < splitPath.length; i += 1) {
+        const pathSegment = splitPath[i]
+        if (pathSegment.startsWith(destPath)) {
+            splitPath[i] = newPathSegment
+            hadExistingPath = true
+            break
+        }
+    }
+
+    if (!hadExistingPath) {
+        splitPath.unshift(newPathSegment)
+    }
+
+    return splitPath.join(path.delimiter)
+}
+
+module.exports = getPath
--- a/src/yvm.js
+++ b/src/yvm.js
@@ -7,12 +7,12 @@ const {
 } = require('./util/version')
 const log = require('./common/log')
 
-const withRcFileVersion = action => (maybeVersionArg, ...rest) => {
+// eslint-disable-next-line consistent-return
+const getYarnVersion = (maybeVersionArg, ...rest) => {
     if (maybeVersionArg) {
         if (isValidVersionString(maybeVersionArg)) {
             log(`Using provided version: ${maybeVersionArg}`)
-            action(maybeVersionArg, ...rest)
-            return
+            return [maybeVersionArg, ...rest]
         }
     }
 
@@ -22,12 +22,17 @@ const withRcFileVersion = action => (maybeVersionArg, ...rest) => {
         const version = getRcFileVersion()
         validateVersionString(version)
         log(`Using yarn version: ${version}`)
-        action(version, ...rest)
+        return [version, ...rest]
     } catch (e) {
         log(e.message)
         process.exit(1)
     }
 }
+
+const invalidCommandLog = () =>
+    log(
+        'You need to source yvm to use this command. run `source /usr/local/bin/yvm`',
+    )
 
 /* eslint-disable global-require,prettier/prettier */
 argParser
@@ -47,10 +52,11 @@ argParser
     .command('install [version]')
     .alias('i')
     .description('Install the specified version of Yarn.')
-    .action(withRcFileVersion(version => {
+    .action(maybeVersion => {
+        const [version] = getYarnVersion(maybeVersion)
         const install = require('./commands/install')
         install(version)
-    }))
+    })
 
 argParser
     .command('remove <version>')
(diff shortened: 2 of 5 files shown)


src/yvm.js
```diff
@@ -98,6 +98,18 @@ argParser
         listVersions()
     })
 
+argParser
+    .command('get-path [version]')
+    .description('Internal command: Gets a new PATH string for "yvm use"')
+    .action(version => {
+        if (!version) {
```
jakebolam: Can we look in the configPath for the version too (i.e. version is optional)
```
        // TODO: says whether installed or not
        if (version) {
            const path = getExtractionPath(version)
            log.capturable(path)
            return
        }

        const configVersion = getRcFileVersion()
        if (configVersion) {
            const path = getExtractionPath(configVersion)
            log.capturable(path)
            return
        }
```

src/yvm.js
```diff
@@ -98,6 +98,18 @@ argParser
...
+            log('Version missing')
```
jakebolam: Can we exit non-zero?
```
        log('No yarn version found or supplied')
        process.exit(1)
?
```

src/yvm.sh
```diff
@@ -4,34 +4,13 @@ command=$1
...
-        yvm_err 'version is required'
-        return 3
-    else yvm_echo "$(yvm_version_dir)/v${VERSION}"
+    local PROVIDED_VERSION=${1}
+    NEW_PATH=$(yvm get-path ${PROVIDED_VERSION})
+    if [ -z "${NEW_PATH}" ]; then
+        yvm_err "Could not get new path from yvm"
+    else
+        PATH=${NEW_PATH}
+        yvm_echo "Set yarn version to ${PROVIDED_VERSION}"
```
jakebolam: [nb] Drop the provided version (as it's hard to get from when yvm finds it from the rc file), `Set yarn to path blah`

src/commands/getPath.js
```diff
@@ -0,0 +1,28 @@
+const path = require('path')
+
+const { versionRootPath, yvmPath } = require('../common/utils')
+
+const getPath = (version, rootPath = yvmPath) => {
```
jakebolam: Can we get a unit tests. This happen to me earlier, would be good if we could extend this to somehow now about global yarn, and remove it too:
```
➜  fe-libraries git:(feature/add-jwt) ✗ yvm use
Found config /code/fe-libraries/.yvmrc
Set yarn to /Users/jakebolam/.yvm/versions/v1.7.0
➜  fe-libraries git:(feature/add-jwt) ✗ $PATH
zsh: not a directory: /Users/jakebolam/.yvm/versions/v1.7.0/bin/yarn/bin:/Users/jakebolam/.nvm/versions/node/v8.11.3/bin:/Users/jakebolam/.yarn/bin:/Users/jakebolam/.config/yarn/global/node_modules/.bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
➜  fe-libraries git:(feature/add-jwt) ✗ yarn --version
1.0.2
```

src/yvm.js
```diff
@@ -98,6 +98,20 @@ argParser
         listVersions()
     })
 
+argParser
+    .command('get-path [version]')
+    .description('Internal command: Gets a new PATH string for "yvm use"')
+    .action(withRcFileVersion(version => {
```
jakebolam: Does this work with a provided version from action?

jakebolam: Little confused about this, would it be better to make the code more explicit? Instead of hiding nested functions

francoiscampbell: yup, the same is used in `yvm install`

src/commands/getPath.js
francoiscampbell: yeah, i'm thinking a shell test or two in addition to unit tests for each command

src/yvm.sh
```diff
@@ -4,34 +4,12 @@ command=$1
...
-
-yvm_version_path() {
-    local VERSION
-    VERSION="${1-}"
-    if [ -z "${VERSION}" ]; then
-        yvm_err 'version is required'
-        return 3
-    else yvm_echo "$(yvm_version_dir)/v${VERSION}"
+    local PROVIDED_VERSION=${1}
+    NEW_PATH=$(yvm get-path ${PROVIDED_VERSION})
```
jakebolam: `yvm` ? does this work?

francoiscampbell: it does, but i can change it to use node directly

src/yvm.js
francoiscampbell: `withRcFileVersion` is actually kinda heavy:

```
const withRcFileVersion = action => (maybeVersionArg, ...rest) => {
    if (maybeVersionArg) {
        if (isValidVersionString(maybeVersionArg)) {
            log(`Using provided version: ${maybeVersionArg}`)
            action(maybeVersionArg, ...rest)
            return
        }
    }

    rest.unshift(maybeVersionArg)

    try {
        const version = getRcFileVersion()
        validateVersionString(version)
        log(`Using yarn version: ${version}`)
        action(version, ...rest)
    } catch (e) {
        log(e.message)
        process.exit(1)
    }
}
```

francoiscampbell: it has to support extra args, etc

src/yvm.sh
dat2: 🐔 and 🥚

PR #125, opened 2018-07-27
Re-enable auto-install on `yvm use`

francoiscampbell:
## Description
That functionality broke in #118

### DevQA Prep

- `make install-watch`
- Open new terminal

### DevQA Steps
- `yvm use <a version that's not installed>`
- Should auto-install and switch to it


brandonbaksh:
QA ✅

PR #142, opened 2018-11-19
Automatically use the specified version of yarn after installing

aeldamaty:
## Description

- yvm will not automatically use the yarn version specified after installing using `yvm install [version]`

## DevQA
- Run `yvm install [version of your choice]`
- It should set that same version as the current yarn version. 
- Can confirm using `yvm which`

### DIFF
--- a/src/yvm.sh
+++ b/src/yvm.sh
@@ -29,16 +29,18 @@ save_last_used_yarn_version() {
 
 yvm_use() {
     local PROVIDED_VERSION=${1}
-    echo ${PROVIDED_VERSION}
     NEW_PATH=$(yvm_call_node_script get-new-path ${PROVIDED_VERSION})
-    echo ${NEW_PATH}
     if [ -z "${NEW_PATH}" ]; then
         yvm_err "Could not get new path from yvm"
     else
         PATH=${NEW_PATH}
+        YARN_VERSION="$(yarn --version)"
+        CURRENT_GLOBAL_YARN_VERSION="${LAST_USED_YARN_VERSION:-}"
         yvm_echo "Now using yarn version $(yarn --version)"
+        if [ ! -e .yvmrc ] && [ "$YARN_VERSION" != "$CURRENT_GLOBAL_YARN_VERSION" ]; then
+            save_last_used_yarn_version ${1}
+        fi
     fi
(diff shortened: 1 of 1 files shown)


brandonbaksh:
QA:
- When opening a new shell the it appears yvm gets removed from the PATH

STR:
- yvm install 1.2.0
- echo $PATH
at this point i can see that yarn 1.2.0 was selected as my default version
- close shell
- reopen shell
- echo $PATH
at this point yvm no longer exists in my PATH which is problematic

brandonbaksh:
QA LGTM ✅

src/yvm.sh
```diff
@@ -2,6 +2,30 @@
...
 YVM_DIR=${YVM_DIR-"${HOME}/.yvm"}
+export_yvm_dir_string="export YVM_DIR=${YVM_DIR}"
+
+save_last_used_yarn_version() {
+    yarn_version="${1}"
+    if [ -z "$yarn_version" ]; then
+        yarn_version="$(yarn --version)"
+    fi
+
+    if [ -z "$yarn_version" ]; then
```
brandonbaksh: What does this block do?

src/yvm.sh
```diff
@@ -2,6 +2,30 @@
...
+        return
+    fi
+    
+    if ! grep -q "LAST_USED_YARN_VERSION" ~/.zshrc; then
+        sed -i '' 's/.*YVM_DIR.*/export LAST_USED_YARN_VERSION='"${yarn_version}"'\'$'\n&/g' ~/.zshrc
```
brandonbaksh: will this create a .zshrc automatically? What if the user doesn't use zsh, we wouldnt want to automatically create a file for them.

src/yvm.sh
aeldamaty: It's just an extra check to return if no last version of yarn was passed in and yarn wasn't actually added to the path for whatever reason (hence yarn --version would fail/return nothing).

src/yvm.sh
aeldamaty: I believe so. Fixing now.

src/yvm.sh
brandonbaksh: why not use add that as an else statement part of the block on line 9?

PR #163, opened 2018-11-22
feat: shim yarn commands

### LINKED ISSUE #136: Automatically use .yvmrc upon entering a directory
When working with multiple projects that utilize different versions of yarn, I always have to type `yvm use` in order to have yvm set the correct version of yarn that my project has. 

When I switch into a directory that already has a `.yvmrc` file, I would like yvm to automatically run `yvm use` so that I don't have to do it manually.

aeldamaty:
## Description

Should resolve #136. All yarn commands will now be shimmed. YVM will look for a `.yvmrc` file to see which version of yarn should be used. If none are found, it will used the last globally set default version of yarn.

### DevQA Prep

- Run `make install`
- Open a new terminal session. `echo $PATH`. You should see the shim directory at the front of your path.

### DevQA Steps
- Run any yarn command. If you do not have a default yarn version set, this will be executed using the `default` `stable` version.
- Switch into any folder with a .yvmrc file with a yarn version that is not installed. Run a yarn command. This will download and install the yarn version before running the command.

### DIFF
--- a/src/yvm.fish
+++ b/src/yvm.fish
@@ -40,11 +40,7 @@ function yvm
             yvm_err "%s\n" "Please ensure your YVM env variables and sourcing are set below sourcing node/nvm in your fish config file"
             exit 1
         end
-
-        set -U DEFAULT_YARN_VERSION (yvm_call_node_script get-default-version 2>/dev/null)
-        if [ "x" != "x$DEFAULT_YARN_VERSION" ]
-            yvm_use > /dev/null
-        end
+        set -U fish_user_paths "$YVM_DIR/shim" $fish_user_paths
     end
 
     if [ "$command" = "use" ]
--- a/src/yvm.fish
+++ b/src/yvm.fish
@@ -15,7 +15,7 @@ function yvm
         if [ -z "$NEW_FISH_USER_PATHS" ]
             yvm_err "Could not get new path from yvm"
         else
-            set -U fish_user_paths $NEW_FISH_USER_PATHS
+            set -U fish_user_paths (string split ' ' -- $NEW_FISH_USER_PATHS)
             set -l new_version (yarn --version)
             yvm_echo "Now using yarn version $new_version"
         end
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
 
 <!-- OVERVIEW-DOCS:START -->
 
-Pesky yarn versions got you down? Automatically and easilly manage those versions.
+Pesky yarn versions got you down? Automatically and easily manage those versions.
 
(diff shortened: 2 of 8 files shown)


mcataford:
@aeldamaty QA steps unclear, can you rephrase/rewrite and run through them?

src/shim/yarn
```diff
@@ -0,0 +1,40 @@
+set -e
```
jakebolam: Hash bang?

jakebolam:
@aeldamaty what is the motivation behind this?

You'll notice there's some been some PR's to assist in this effort:
https://github.com/tophat/yvm/pull/118

jakebolam:
Have put up #189 which removes the need for this

aeldamaty:
@jakebolam so the motivation behind this is to get rid of the need to type in `yvm use`. It shims any yarn command and automatically appends the correct version to the PATH. 

It will probably need some refactoring once your global version of yarn PR goes in, but essentially it would work as follows. 

- Any yarn command you entered would be shimmed
- It would determine the version of yarn to use by looking for a .yvmrc file
- If no .yvmrc file exists, it will use the global version
- If none set it would error out. 

Also, the way it currently works, if the user ever types in `yvm use` they have essentially taken over control for just this session and it will use the version of yarn they specify at all times without shimming.

src/shim/yarn
```diff
@@ -0,0 +1,40 @@
...
+
+program="${0##*/}"
```
jakebolam: Can we get tests for his file in CI somehow

src/shim/yarn
```diff
@@ -0,0 +1,40 @@
...
+if [ $current_path = *.yvm/versions/v* ]; then
+    yarn_version="$(yarn --version)"
+    yarn_version_base_directory="${yvm_dir}/versions/v${yarn_version}/"
+    yarn_directory="${yarn_version_base_directory}bin/"
+    exec -a "$yarn_directory" "$program" "$@"
+    return
+fi
+
+if [ "$program" = "yarn" ]; then
+    yarn_version="${LAST_USED_YARN_VERSION:-}"
```
jakebolam: Can this logic be moved to js?

jakebolam:
@aeldamaty I like the motivation to solve the cd'ing into directories etc by shimming the yarn command. Once the global version PR is in, lets totally figure out this logic

mcataford:
@aeldamaty Let's revive and drive this one home 🚀

mcataford:
@aeldamaty Any news on this? :rocket:

src/yvm.fish
```diff
@@ -15,7 +15,7 @@ function yvm
         if [ -z "$NEW_FISH_USER_PATHS" ]
             yvm_err "Could not get new path from yvm"
         else
-            set -U fish_user_paths $NEW_FISH_USER_PATHS
+            set -U fish_user_paths (string split ' ' -- $NEW_FISH_USER_PATHS)
```
iamogbz: This is to fix fish not splitting on spaces but [on new lines](https://fishshell.com/docs/current/faq.html)

PR #197, opened 2018-12-04
bugfix: duplicate help command

jakebolam:
## Description
Before:
```
    set-default <version>     Sets the default fallback yarn version, if not supplied and no .yvmrc found
    get-default-version       Gets the default set yarn version
    version                   Outputs the version of yvm that is installed
    update-self               Updates yvm to the latest version
    help                      Show help text
    help [cmd]                display help for [cmd]
```
screen shot 2018-12-04 at 6 21 11 pm

After:
```
    set-default <version>     Sets the default fallback yarn version, if not supplied and no .yvmrc found
    get-default-version       Gets the default set yarn version
    version                   Outputs the version of yvm that is installed
    update-self               Updates yvm to the latest version
    help                      Show help text
```

### DIFF
--- a/src/yvm.js
+++ b/src/yvm.js
@@ -135,10 +135,5 @@ argParser
     .description('Updates yvm to the latest version')
     .action(invalidCommandLog);
 
-argParser
-    .command('help')
-    .description('Show help text')
-    .action(() => argParser.outputHelp());
-
 /* eslint-enable global-require,prettier/prettier */
 argParser.parse(process.argv)


brandonbaksh:
@jakebolam I guess we have "two" usages for the help command.
1) yvm
 - This should output the list of commands (just like the hep page)

2) yvm help
- This should do the same thing as 1)

I think with your changes it invalidates 2).

See the screenshot below:
screen shot 2018-12-05 at 10 33 19 am


Notice it says `Invalid help command` for `yvm help`

jakebolam:
@brandonbaksh that `yvm help` invalid command is a tricky one
as for the confusing output of `help [cmd]` this is recently coming from a new version fo commander.
Looking at there docs/github there is no current way to disabled it. May need to put a PR up on there side. (Added an issue: https://github.com/tj/commander.js/issues/892)

PR #199, opened 2018-12-05
fix: make sure greenkeeper can generate lockfiles

jakebolam:
greenkeeper is having trouble with lockfiles


Makefile
```diff
@@ -7,7 +7,11 @@ ifdef CI
     JEST_ENV_VARIABLES=JEST_SUITE_NAME=yvm JEST_JUNIT_OUTPUT=$(TEST_REPORTS_DIR)/tests/jest.junit.xml
     JEST_ARGS=--ci --maxWorkers=2 --reporters jest-junit
     WEBPACK_ARGS=
+ifneq (,$(findstring greenkeeper,$(CIRCLE_BRANCH)))
```
jakebolam: if has greenkeeper in the branch name, then don't lock the file

jakebolam:
This is not the case. Reverting. 
See: https://github.com/greenkeeperio/greenkeeper/issues/981

Issue #207, opened 2018-12-09
test: automated testing of different shells in CI


jakebolam:
**Is your feature request related to a problem? Please describe.**
Ensure we automatically test common shells in CI

**Describe the solution you'd like**
Test install, and run

**Describe alternatives you've considered**

**Additional context**
As the project gains more maturity, we want more confidence in our releases 
- See [bats-core](/bats-core/bats-core)

Issue #111, opened 2018-07-04
Idea: use exec instead of __non_webpack_require__


dat2:
Instead of requiring the yarn bin js file, lets `exec` the yarn command.

If yarn ever changes their entry point to a shell script instead of a js file, we will suddenly stop working 😞

msrose:
https://nodejs.org/dist/latest-v10.x/docs/api/child_process.html#child_process_child_process_exec_command_options_callback

jakebolam:
in #213, released in v0.9.3: https://github.com/tophat/yvm/releases/tag/v0.9.3

Issue #134, opened 2018-10-16
Shimming the yarn command


dat2:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to have the `yarn` command automatically get `use`d depending on the folder i'm in! I get frustrated when I have to `yvm use` every time I open a new terminal.

**Describe the solution you'd like**
Similar to `pyenv` when you `cd` into a directory that has the `.yvmrc` file in it, and the yarn version specified is installed, shim the `yarn` command to be automatically `use`d.

Example:

`my-project-1` uses yarn `1.7.0`
```
➜ cd my-project-1
➜ cat .yvmrc
1.7.0
➜ yarn --version
1.7.0
```

`my-project-2` uses yarn `1.2.0`
```
➜ cd ../my-project-2
➜ cat .yvmrc
1.2.0
➜ yarn --version
1.2.0
```

I don't have yarn installed globally
```
➜ cd ..
➜ yarn --version
zsh: command not found: yarn
```

**Describe alternatives you've considered**
nothing 🙃 

**Additional context**

brandonbaksh:
Please refer to #136

PR #213, opened 2018-12-09
feat: spawn shell to use yvm exec

### LINKED ISSUE #214: yvm exec with no args crashes
**Describe the bug**
yvm exec with no args crashes.

**To Reproduce**
1. `yvm exec`

jakebolam:
## Description
Addresses #111. Fixes #214. First step towards #207 

Will need the switch to node_modules PR in first.


src/yvm-exec.js
```diff
@@ -9,32 +9,26 @@ const { yvmPath } = require('./util/utils')
...
-        ? __non_webpack_require__
-        : require
-/* eslint-enable no-undef, camelcase */
-
 const runYarn = (version, extraArgs, rootPath) => {
     // first two arguments are filler args [node version, yarn version]
     process.argv = ['', ''].concat(extraArgs)
-    log.info(`yarn ${extraArgs.join(' ')}`)
-    execFile(path.resolve(getYarnPath(version, rootPath), 'bin/yarn.js'))
+    const filePath = path.resolve(getYarnPath(version, rootPath), 'bin/yarn.js')
```
francoiscampbell: in order to address #111, shouldn't we exec the main `yarn` command instead of the js?

jakebolam: Planning to shim it possibly, does this affect it? #134

francoiscampbell: good point, we can cross that bridge when we get there

PR #216, opened 2018-12-10
deploy: only zip production deps in production build

jakebolam:
```
du -hs node_modules
191M	node_modules
du -hs node_modules_production
46M	node_modules_production
```


.circleci/config.yml
```diff
@@ -52,6 +52,7 @@ jobs:
     working_directory: ~/repo
     steps:
       - prep_env
+      - run: rm -rf node_modules && yarn install --production --frozen-lockfile
```
jakebolam: need to actually do a step in between with build-production (as we need dev deeps to build some parts)

PR #218, opened 2018-12-10
Prevent yvm.zip from including itself; force overwrite in dev install

francoiscampbell:
## Description
- A bug in https://www.npmjs.com/package/webpack-zip-files-plugin makes it include the output file in itself if the file is created in the directory being zipped up
- This PR changes the output path of `yvm.zip` to be `./artifacts/yvm.zip` instead of `./artifacts/webpack_build/yvm.zip`
- This PR also forces `unzip` to overwrite any files when installing using `make install`. It was already doing that when using the `curl | bash` method

### DevQA Steps
- Running `make install` more than once should not prompt for overwrite, it should just do it.
- The output `yvm.zip` should be located in the `artiacts` directory
- Extracting or listing the contents of `yvm.zip` should not show another `yvm.zip` within

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -11,7 +11,6 @@ trap 'err_report $LINENO' ERR
 USE_LOCAL=${USE_LOCAL-false}
 
 release_api_url="https://api.github.com/repos/tophat/yvm/releases/latest"
-artifacts_dir="artifacts/webpack_build"
 
 YVM_DIR=${YVM_INSTALL_DIR-"${HOME}/.yvm"}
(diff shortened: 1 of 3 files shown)


scripts/install.sh
```diff
@@ -11,7 +11,6 @@ trap 'err_report $LINENO' ERR
 USE_LOCAL=${USE_LOCAL-false}
 
 release_api_url="https://api.github.com/repos/tophat/yvm/releases/latest"
-artifacts_dir="artifacts/webpack_build"
```
francoiscampbell: this was never used

PR #220, opened 2018-12-10
Fix bash error: "local can only be used in a function"

msrose:
When opening a new shell in bash, yvm gives the following error:

```
-bash: local: can only be used in a function
```

I believe this fixes it. Still untested.

UPDATE: modified files locally and this does indeed fix the error.

### DIFF
--- a/src/yvm.sh
+++ b/src/yvm.sh
@@ -55,7 +55,7 @@ if [ ${interactive} = 1 ]; then
     yvm() {
         yvm_ 'function' $@
     }
-    local DEFAULT_YARN_VERSION=$(yvm_call_node_script get-default-version)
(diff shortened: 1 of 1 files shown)


jakebolam:
Great thanks @msrose more reason for #207 

@msrose what version of bash is your shell? What shell was it spawned from?

jakebolam:
@msrose in release v0.9.8 https://github.com/tophat/yvm/releases/tag/v0.9.8

msrose:
```
bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
Copyright (C) 2007 Free Software Foundation, Inc.
```

PR #238, opened 2018-12-12
Add Global default while printing versions

### LINKED ISSUE #193: yvm list should show the default global version
`yvm list` should show the default global version

The default global version to appear in the list

n/a

**Additional context**
n/a

yashshah:
## Description

Fix #193 
When printing the version of the yvm installed/available, it should show the default global version
It also moves the path logic out of utils.js to another helper file path.js so that no cyclic dependency is introduced.

### DevQA Steps
- `make Install`
- `yvm list`

### Comments:
image

image


jakebolam:
Thanks @yashshah

PR #241, opened 2018-12-13
Auto infer version

### LINKED ISSUE #194: auto infer version (instead of expecting in exact format)
There's currently a long list of non-supported versions where we can easily infer the correct version. We could automatically infer the correct version to make yvm more user friendly

Infer the correct version, and move a bunch of these invalid version cases, to the valid version cases:
https://github.com/tophat/yvm/blob/master/test/util/version.test.js#L38

Not doing it

**Additional context**

yashshah:
## Description

Add an option to infer range of version string which is prefixed with `v`. For example: `v1.0.0`  
Fix #194 

### DevQA Steps
`make install`
`yvm i v1.2.0`
`yvm use v1.2.0`

### Comments:

Right now regex is quite liberal with the text around the actual number, should we limit that only have optional prefix `v` before the version number?


jakebolam:
@yashshah this looks great. CR looks good. Want to resolve merge conflicts and then I'll QA?

jakebolam:
FYI valid version with this can be thing slike this too 🤔 `yvm use v1.9.4asdsadasda`

yashshah:
You're right @jakebolam. Good catch! Something like `2.0.0-beta.3` can be valid version number. We also have `1.1.0-exp.2` in list-remote which we should also consider. I will make the change for this.

yashshah:
@jakebolam I've fixed the conflicts. Can you please review this now?

src/yvm.js
```diff
@@ -117,19 +120,22 @@ argParser
...
     .action(version => {
-        if (!isValidVersionString(version)) {
+        const parsedVersionString = getValidVersionString(version)
+        if (!parsedVersionString) {
             log('Error: Invalid version string supplied.')
             process.exit(1)
             return
         }
-        if (setDefaultVersion({version})) {
+        if (setDefaultVersion({ version })) {
```
jakebolam: Should this be the parsed version string?

jakebolam:
@yashshah 1 comment. LGTM tho.

yashshah:
@jakebolam Nice catch, I've made the changes. Can you please review again? Thanks!

src/yvm.js
```diff
@@ -48,88 +47,99 @@ argParser
...
     .action(version => {
-        if (!isValidVersionString(version)) {
+        const parsedVersionString = getValidVersionString(version)
+        if (!parsedVersionString) {
             log('Error: Invalid version string supplied.')
             process.exit(1)
             return
         }
-        if (setDefaultVersion({version})) {
+        if (setDefaultVersion({ version: parsedVersionString })) {
```
jakebolam: @yashshah with named params we need to do this one like this :)

jakebolam:
@yashshah feel free to merge this :)

Issue #101, opened 2018-06-26
Highlight currently active Yarn version in `yvm list` and `yvm list-remote`


umar-tophat:
**Is your feature request related to a problem? Please describe.**
If I have multiple versions of Yarn installed and I want to switch from one to another, I have to first run `yvm which` to determine which version I am currently using. Then I have to remember this version, and run `yvm list` to see what versions are installed. This requires more cognitive overhead than should be required for a tool like this.

**Describe the solution you'd like**
When you run `yvm list` or `yvm list-remote`, the currently active version should be highlighted with a different color AND an arrow should be pointing to it like in nvm (see screenshot below).

**Describe alternatives you've considered**
An alternate solution is to simply add a line at the beginning or end of the version output that states what version is currently being used. This approach is not the best because there is a lot of info being printed on the screen in these commands, so it is likely to be missed.

**Additional context**

![nvm list output](https://user-images.githubusercontent.com/38886386/41938315-bcc533a4-7960-11e8-8287-8ed7d80838a4.png)

sanchitgera:
Fixed ls-remote in #248

PR #248, opened 2018-12-17
Highlight active Yarn version when listing remote versions

sanchitgera:
## Description
Highlight the version in use when listing remote versions. For #101 

### DevQA Steps
- Create a directory with a .yvmrc file 
- Run `yvm use` 
- Run `yvm ls-remote`
- See that the active version is highlighted in green


src/commands/list.js
```diff
@@ -1,9 +1,8 @@
 const fs = require('fs')
-const { exec } = require('shelljs')
```
jakebolam: 👍

src/commands/list.js
```diff
@@ -1,9 +1,8 @@
...
 
 const log = require('../util/log')
 const { stripVersionPrefix, versionRootPath } = require('../util/utils')
-const { printVersions } = require('../util/version')
+const { getVersionInUse, printVersions } = require('../util/version')
```
jakebolam: 👍

jakebolam:
@sanchitgera CR looks good, I'll qa this tomorrow

Issue #247, opened 2018-12-16
Add bundlewatch to project to prevent sharp or accidental increases in release size


jakebolam:
**Is your feature request related to a problem? Please describe.**
We want to keep the release as small as possible

**Describe the solution you'd like**
Add bundlewatch to have visibility into branches increasing the bundle/release size

**Describe alternatives you've considered**

**Additional context**
Could help prevent issues such as #245

PR #252, opened 2018-12-17
ci: add bundlewatch

jakebolam:
## Description
Protect against accidental bumps in release size see #247 

## DevQA
CI Passes


jakebolam:
@torinthiel Could you review?

sanchitgera:
CR LGTM, Will QA in a bit

Issue #255, opened 2018-12-18
Exceeding GitHub Rate Limit - API rate limit exceeded for xxx


sanchitgera:
**Describe the bug**
Installing yvm using the script fails sporadically. We rely on the GitHub API to find the latest release of YVM and then download it. 
https://github.com/tophat/yvm/blob/27d51d0b4bae565a2f79cb45166412b53b2b28fd/scripts/install.sh#L13
But GitHub's API restricts us to about 50 or requests every hour. If you have other apps on your system that use GitHub's API, then those requests will count against this limit as well. 
This is a problem when you're developing locally and end up exceeding the limit. You then get locked out for the next hour or so. 

Also had this show up during our CI process:
https://circleci.com/gh/tophat/yvm/1791?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

**To Reproduce**
Hammer the system by calling `yvm ls-remote` or run `install.sh` lots of times 

**Expected behavior**
Not sure how we want to handle this issue, or if we should worry about it until we've had some people reporting this as an actual concern. 

**Additional context**
We could also switch to using read-only access credentials instead of making unauthenticated API requests. This increases the API limit from ~50 to ~5000.

jakebolam:
Workaround coming in #256 to support specific version/skipping github api for install.

Will still need to do for `yvm ls-remote`

shawndrape:
If we wanted to go the extra mile here, we could detect the rate limiting, throw up a useful message and/or prompt for Github credentials to retry with the authenticated call.

jakebolam:
Could add docs like all-contributors does: https://github.com/mntnr/name-your-contributors/blob/master/readme.md#api-limits-and-setting-up-a-github-token
Or something like #267

jakebolam:
This is actually super bad. Just hit the rate limit on `yvm install` as we validate if the remote version is correct.

jakebolam:
This is affecting:
- `yvm exec` when version not instaled
- Shell insilization when version is not installed `yvm get-new-path` calls ensure
- `install.sh`: 
- `yvm list-remote`
- `yvm update-self`

`install.sh` error crash:
```
yvm update-self
Querying github release API to determine latest version
[eval]:1
var stdin = fs.readFileSync(0).toString(); var json = JSON.parse(stdin); console.log(json.assets[0].browser_download_url)
                                                                                                ^

TypeError: Cannot read property '0' of undefined
    at [eval]:1:97
[...]

PR #256, opened 2018-12-19
feat: support install without GitHub api

jakebolam:
## Description
Provides workaround for #255

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,11 @@ Execute the following in your terminal:
 curl -fsSL https://raw.githubusercontent.com/tophat/yvm/master/scripts/install.sh | bash
 ```
 
+Or to install a specific version:
+```bash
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -59,6 +59,11 @@ Execute the following in your terminal:
 curl -fsSL https://raw.githubusercontent.com/tophat/yvm/master/scripts/install.sh | bash
 ```
 
+Or to install a specific version:
```
sanchitgera: I don't know if we should make this API public until we've finalized a solution for #255 🤷‍♂️

jakebolam: I think installing a specific version is useful, even if not to fix #255. Been meaning to add something for this 👍 

Will still need additional work on both fronts. e.g. should document pinning a specific version of install.sh too

PR #258, opened 2018-12-19
Less verbose initialization

### LINKED ISSUE #226: Should not output anything on starting new shell
After installation, when a user opens new shell, `yvm` displays the default `yarn` version. After some time this is getting unnecessary, and in some cases, when the executed shell is supposed to be non-interactive, may be problematic, as it may influence another program that is using the created shell.

The message about currently selected version should be suppressed when initially loading `yvm`.

Current situation: Unnecessary, possibly harmful message.

**Additional context**

torinthiel:
Make initialization (sourcing from bashrc/zshrc) output only error
messages, silence default messages from yvm use.

## Description
Changed `yvm.sh` so that the initialization is quiet, displaying only error messages.

### DevQA Steps
Make sure that initialization is quiet (no message about currently used version), however error messages (e.g. about missing node version) are still present. The message should still be displayed if `yvm use` is used directly.
Please test both `bash` and `zsh`, especially as I don't use `zsh` myself.

### Comments:
Fixes #226

### DIFF
--- a/src/yvm.sh
+++ b/src/yvm.sh
@@ -55,16 +55,13 @@ if [ ${interactive} = 1 ]; then
     yvm() {
         yvm_ 'function' $@
     }
-    DEFAULT_YARN_VERSION=$(yvm_call_node_script get-default-version)
-    if [ "x" = "x${DEFAULT_YARN_VERSION}" ]; then
-        yvm_echo "Use yvm set-default <version>"
-    else
-        if ! type "node" > /dev/null; then
-            yvm_echo "YVM Could not automatically set yarn version."
-            yvm_echo "Please ensure your YVM env variables and sourcing are set below sourcing node/nvm in your .zshrc or .bashrc"
-        else
-            yvm_use
-        fi
+    if ! type "node" > /dev/null; then
+        yvm_err "YVM Could not find node executable."
+        yvm_err "Please ensure your YVM env variables and sourcing are set below sourcing node/nvm in your .zshrc or .bashrc"
+    fi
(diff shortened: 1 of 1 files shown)


src/yvm.sh
```diff
@@ -55,16 +55,13 @@ if [ ${interactive} = 1 ]; then
...
-        else
-            yvm_use
-        fi
+    if ! type "node" > /dev/null; then
+        yvm_err "YVM Could not find node executable."
+        yvm_err "Please ensure your YVM env variables and sourcing are set below sourcing node/nvm in your .zshrc or .bashrc"
+    fi
+    DEFAULT_YARN_VERSION=$(yvm_call_node_script get-default-version 2>/dev/null)
+    if [ "x" != "x${DEFAULT_YARN_VERSION}" ]; then
+        yvm_use > /dev/null
```
jakebolam: Thoughts on always calling yvm-use. With no concept of DEFAULT_YARN_VERSION here.

We could improve yvm-use to alert when no default version is set (when called outside of this init context) or perhaps provide a built-in fallback version. I.e. if it can't find a version it will use yarn 1.7.2.

torinthiel: Currently `yvm use` displays warning message on stderr when there's no default version set.

I can change the code to call yvm_use unconditionally, but this will mean users receive a 'nag' with every shell until they decide to use `yvm set-default`. Or we can put this on hold until #261.

torinthiel:
@jakebolam I guess your approval means 'Let's deal with unconditional use later'. I'll give you some time to proove me wrong, and merge later ;)

PR #259, opened 2018-12-19
Don't create shell init scripts if not needed

### LINKED ISSUE #228: Installation gives harmless error message when user has no ~/.zshrc or ~/.bashrc
**Describe the bug**
When user has no profile file for the shell (`~/.zshrc` or `~/.bashrc`), installer displays a harmless error message e.g. `grep: /home/torinthiel/.zshrc: No such file or directory` and creates a new file

**To Reproduce**
1. Ensure that you have no `~/.zshrc`
2. Install `yvm`
3. See error message during installation. Check that installation works correctly, but a minimal `~/.zshrc`, containing only instructions for `yvm` is created.

**Expected behavior**
The error message should not appear. I have no idea whether we should create a new file for that shell if user had none.

**Desktop (please complete the following information):**
 - OS: linux
 - YVM Version: 0.9.10
 - Shell: bash/zsh
 - Node version: 10.14.1

**Additional context**
Note that this will get more and more complex as we add support for more shells - e.g. with #116.

torinthiel:
The scripts ~/.zshrc and ~/.bashrc should only be modified if already
existing, otherwise assume user does not need those and do not create
them.

Also fixes a problem when nonexisting file would cause installation
script to display error message during installation.

## Description
Changed installation script to check for existence of `~/.zshrc` / `~/.bashrc` before invoking `grep` or modifying them.

### DevQA Steps
Check installation when `~/.zshrc` / `~/.bashrc` are or are not present in user's home directory.

### Comments:
Fixes #228

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -54,28 +54,32 @@ fi
 
 chmod +x ${sh_install_path}
 
-added_newline=0
-if ! grep -q "${export_yvm_dir_string}" ~/.zshrc; then
-    echo '' >> ~/.zshrc
-    echo ${export_yvm_dir_string} >> ~/.zshrc
-    added_newline=1
+if [ -f ~/.zshrc ]; then
+    added_newline=0
+    if ! grep -q "${export_yvm_dir_string}" ~/.zshrc; then
+        echo '' >> ~/.zshrc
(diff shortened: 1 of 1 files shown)


jakebolam:
:+1:

PR #263, opened 2018-12-20
ci: pin yvm version used in pipeline

jakebolam:
Fixes ci flakes where github quota is being hit

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ commands:
           key: yvm-{{ .Branch }}-{{ checksum "yarn.lock" }}
       - run:
           name: 'Install yvm release version'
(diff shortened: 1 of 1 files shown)


jakebolam:
@sanchitgera noticed your branch hitting this too. Mind approving?

sanchitgera:
Why is CI relying on this script though? And if it is, wouldn't it always want to get the latest version of YVM?

jakebolam:
It's basically dog fooding yvm.

We use yvm's previous release to setup the pipeline to build the new release of yvm.

Pinning will get us around the quota (while not ideal) i'm currently looking at other alternatives too

jakebolam:
The dog fooding actually caught an issue: #268

PR #273, opened 2018-12-26
feat: fish shell support

### LINKED ISSUE #116: Support Fish shell
To use `yvm use`, I have to source yvm within my shell environment. However, because yvm shell syntax is not compatible with fish shell,
sourcing will not work properly.

```
> source /usr/local/bin/yvm
/usr/local/bin/yvm (line 25): Unsupported use of '&&'. In fish, please use 'COMMAND; and COMMAND'.
    [ -n "${1-}" ] && [ -x "$(yvm_version_path "$1" 2> /dev/null)"/bin/yarn ]
                    ^
from sourcing file /usr/local/bin/yvm
        called on standard input

source: Error while reading file '/usr/local/bin/yvm'
```

Provide additional shell script for sourcing yvm in fish shell. 

`virtualenv` uses a similar approach. In its environment directory, has separate activate scripts for bash, fish shell, and windows CMD.

**Additional context**

eastenluis:
## Description
This resolves #116. When fish shell executes `yvm use`, it runs the bash script in bash shell, but not interactive mode. I consider running fish shell command within bash script, but every time the command is called, it boosts a new fish shell which adds a lot of overhead from the fish profile. After a few more trials, I decided to write a complete new fish shell script `yvm.fish`, which supports the same functionalities as `yvm.sh`. The main differences are:
- The added yvm path is set to `$fish_user_paths` instead because fish shell does not allow change `PATH` directly.
- Use `getNewFishUserPaths` as fish shell's path is an array, requiring a slightly different format.

The `yvm` comamnds defined in `yvm.fish` will be sourced within fish profile (`~/.config/fish/config.fish`), and it will take higher priority over functions sourced in `yvm.sh`.

The downside of this approach is that we need to maintain two copies of yvm shell script. But hopefully it will be ok since they are thin layers over the yvm logics.

### DevQA Prep

- Please `make install` to ensure `yvm.fish` will be copied to the YVM_DIR.
- Please use a fish shell to test this feature.

### DevQA Steps

Install YVM:
- Open a fish shell and run `make install`. 
- In another fish shell, run `type yvm`. It should say that "yvm is a function with definition" defined in the `yvm.fish` file.

Test each yvm command is working properly, including:
- install
- remove
- exec
- use
- list
- list-remote
- get-default-version
- set-default
- version
- update-self
- help

### Comments

Still misses a test for `getNewFishUserPaths`. Create the PR for preliminary review.


src/yvm.js
```diff
@@ -115,6 +115,19 @@ argParser
         });
     });
 
+argParser
+    .command('get-new-fish-user-path [version]')
```
jakebolam: Same as above:
Could we use 
`yvm_call_node_script get-new-user-path --shell=fish` instead?

jakebolam:
@eastenluis Thank you! Looks pretty good, one suggestion let me know!

src/yvm.fish
```diff
@@ -0,0 +1,65 @@
...
+    if [ "$command" = "use" ]
+        if test (count $argv) -gt 1
+            yvm_use $argv[2]
+        else
+            yvm_use
+        end
+    else if [ "$command" = "update-self" ]
+        curl -fsSL https://raw.githubusercontent.com/tophat/yvm/master/scripts/install.sh | YVM_INSTALL_DIR=${YVM_DIR} bash
+    else if [ "$command" = "init-fish" ]
+        yvm_init_fish
```
jakebolam: is this to scope the function?

src/yvm.fish
```diff
@@ -0,0 +1,65 @@
...
+
+    if [ "$command" = "use" ]
+        if test (count $argv) -gt 1
+            yvm_use $argv[2]
+        else
+            yvm_use
+        end
+    else if [ "$command" = "update-self" ]
+        curl -fsSL https://raw.githubusercontent.com/tophat/yvm/master/scripts/install.sh | YVM_INSTALL_DIR=${YVM_DIR} bash
+    else if [ "$command" = "init-fish" ]
```
jakebolam: is this to scope the function?

scripts/install.sh
```diff
@@ -82,5 +82,23 @@ if [ -f ~/.bashrc ]; then
     fi
 fi
 
+if [ -f ~/.config/fish/config.fish ]; then
```
jakebolam: [nb] making a reusable function for the shell adds

src/yvm.fish
eastenluis: I did the init part outside the `yvm` function at first, and then I realized I needed to re-use function like `yvm_call_node_script`. I can't directly call them outside `yvm` function scope, and if I defined that function outside function scope, it will be immediately sourced as a command available to the whole fish session (b/c all functions in yvm shell script are public by default). 

Therefore I wrap the logic into `yvm init-fish` instead so I can re-use the function. But It's not absolutely necessary.

src/yvm.fish
```diff
@@ -0,0 +1,65 @@
...
+function yvm
+    set command $argv[1]
+
+    function yvm_use
+        if test (count $argv) -gt 0
+            set PROVIDED_VERSION $argv[1]
+        end
+
+        begin
+            set NEW_FISH_USER_PATHS (yvm_call_node_script get-new-fish-user-path $PROVIDED_VERSION)
```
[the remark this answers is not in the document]
eastenluis: Sounds cool. I'll update that and the tests. I will need to update it a bit so that it knows it should fine fish path instead too.

eastenluis:
Updated. Sorry for the delays. 

@iamogbz Could you devQA a bit? I'm not sure how it works on other people's fish shell.

src/yvm.fish
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    if [ "$command" = "use" ]
+        if test (count $argv) -gt 1
+            yvm_use $argv[2]
+        else
+            yvm_use
+        end
+    else if [ "$command" = "update-self" ]
+        curl -fsSL https://raw.githubusercontent.com/tophat/yvm/master/scripts/install.sh | YVM_INSTALL_DIR=${YVM_DIR} bash
```
iamogbz: fishy
```suggestion
        env YVM_INSTALL_DIR=$YVM_DIR curl -fsSL https://raw.githubusercontent.com/tophat/yvm/master/scripts/install.sh | bash
```

eastenluis: Why is it sushi? Is that the only fish emoji you can find? 😂

eastenluis: Updated. Should be working now.

eastenluis:
@iamogbz I think it is in a very similar situation like bash, which it is hard to test the script in an isolated environment, without affecting or affected by the existing fish shell codes.

I think in general the best way is still trying to move more logics into js rather than in the shell scripts...

eastenluis:
@allcontributors[bot] please add @eastenluis for code.

PR #205, opened 2018-12-09
WIP refactor which command

umar-ahmed:
## Description

- Add chalk dependency
- Add error and success coloured log functions
- Extract logic for getting active yarn version into helper function

### DevQA Prep

- Please `make install`, this PR uses a new package

### DevQA Steps

- Run yvm which in the various use cases
	- Verify that errors are shown in red
	- Verify that message for matching version in .yvmrc is shown in green

### Comments:

- This is a WIP
- Replaces https://github.com/tophat/yvm/pull/122

PR #275, opened 2019-01-02
Install archived yarn versions

### LINKED ISSUE #271: Unable to install version `1.1.0-exp.2`
**Describe the bug**
Installing version `1.1.0-exp.2` fails. It is a valid version but currently yvm doesn't allow to install it. 

**To Reproduce**
1. `yvm i 1.1.0-exp.2`
2. See error

**Expected behavior**
It should install version `1.1.0-exp.2`

**Screenshots**
image

iamogbz:
## Description

This should resolve #271.

### DevQA Prep

- Please `make install-watch`

### DevQA Steps

- Should complete successfully and use latest yarn
  - `yvm i -l`
  - `yvm use <LATEST_YARN_VERSION>`
  - `yarn --version`
- Should complete successfully and use installed yarn
  - `yvm i 1.12.0`
  - `yvm use 1.12.0`
  - `yarn --version`
- Should not complete but print appropriate instructions
  - `yvm install 1.3.0`
- Should not complete but print appropriate instructions
  - `yvm i 1.1.0-exp.2`
- Should list all available yarn versions indicating installed versions
  - `yvm ls-remote`

### Comments:

- This is caused by a defective yarn release. https://github.com/yarnpkg/yarn/issues/4828
  - Though yarn can be build from source, `yvm` will not support these broken releases


src/commands/install.js
```diff
@@ -115,33 +128,40 @@ const installVersion = async (version, rootPath = yvmPath) => {
...
         log(
             'You have provided an invalid version number. use "yvm ls-remote" to see valid versions.',
         )
         throw new Error('Invalid version number provided')
     }
 
     log(`Installing yarn v${version} in ${rootPath}`)
-    await downloadVersion(version, rootPath)
-
+    const hasSig = await downloadVersion(version, rootPath, releases[version])
```
jakebolam: [nb] full names `hasSignature`

src/commands/install.js
```diff
@@ -30,10 +32,16 @@ const isVersionInstalled = (version, rootPath) => {
     return fs.existsSync(versionPath)
 }
 
-const downloadVersion = (version, rootPath) => {
+const downloadVersion = async (version, rootPath, releaseInfo) => {
     const url = getUrl(version)
     const filePath = getDownloadPath(version, rootPath)
-    return downloadFile(url, filePath)
+    try {
+        await downloadFile(url, filePath)
```
jakebolam: [nb] not sure if we should use releaseInfo for this too?

src/commands/install.js
```diff
@@ -78,27 +86,32 @@ const verifySignature = async (version, rootPath) => {
...
                     reject(err)
                 } else {
                     log(`Finished extracting yarn version ${version}`)
-                    fs.renameSync(
-                        `${destPath}/yarn-v${version}`,
-                        `${destPath}/v${version}`,
-                    )
-                    fs.unlinkSync(srcPath)
-                    resolve()
+                    const [pkgDir] = fs.readdirSync(tmpPath)
```
jakebolam: Why the change in behaviour here?

src/util/download.js
```diff
@@ -1,17 +1,19 @@
...
-        })
+const downloadFile = (url, filePath) =>
+    new Promise((resolve, reject) => {
+        const handleError = err => reject(err)
+        request
+            .get(url, { headers: { 'user-agent': 'yvm' } })
+            .on('error', handleError)
+            .on('response', r => {
+                const msg = `HTTP ${r.statusCode} - ${r.statusMessage} (${url})`
+                if (r.statusCode === 404) handleError(msg)
```
jakebolam: if not 200 ?

src/commands/install.js
jakebolam: Does release use the mirror download?

src/commands/install.js
```diff
@@ -115,33 +128,40 @@ const installVersion = async (version, rootPath = yvmPath) => {
...
 
-    log('GPG signature validated')
-    return extractYarn(version, rootPath)
+    let continueInstall
+    if (hasSig) {
+        await verifySignature(version, rootPath)
+        log('GPG signature validated')
+        continueInstall = true
+    } else {
+        const keyword = 'yes'
```
jakebolam: Instead of using input. Can we say run `yvm install --ignore-signature` ?
(I understand there are cases we try to auto install, but for these cases I think its fine to fail)

src/util/utils.js
```diff
@@ -32,18 +32,24 @@ const getVersionsFromTags = () => {
                 reject(error)
             } else {
                 const tags = JSON.parse(body)
-                const tagNames = tags.map(tag => tag.name)
-                const versions = tagNames
-                    .map(stripVersionPrefix)
-                    .filter(version => version[0] > 0)
-                resolve(versions)
+                const releases = tags.reduce((accumulator, tag) => {
```
jakebolam: Whats the motivation behind this change?

src/commands/install.js
iamogbz: The `tarball_url` in the cf release info is the src which is much larger. I was thinking use this only as a fallback url.

src/commands/install.js
iamogbz: The extracted file from the fallback `tarball_url` doesn't follow the same naming pattern. However it's still one content folder when extracted.

src/commands/install.js
iamogbz: That probably better fits with the existing pattern 👍

src/util/download.js
iamogbz: It could be a redirect as well. Will update to use the range `2xx` to `3xx`.

src/util/utils.js
iamogbz: Needed the full release info by version to get the fallback `tarball_url`. Preserved the `getVersionsFromTags` fn contract below.

src/commands/install.js
```diff
@@ -78,34 +86,48 @@ const verifySignature = async (version, rootPath) => {
...
                 }
             },
         )
     })
 }
 
-const installVersion = async (version, rootPath = yvmPath) => {
+const buildYarn = srcPath => {
+    const options = { cwd: srcPath, stdio: 'inherit' }
+    execSync('yarn install && yarn run build', options)
```
jakebolam: Installing these versions of yarn need building?

jakebolam: Seems super weird that we need this.

I think we either target the yarn version installed for yvm directly. (We installed yarn we know where it is) then perhaps exit if none installed.

Or maybe we just don't support these versions?

iamogbz: Switched to using `npm` as it comes with `node` which is a requirement

README.md
```diff
@@ -206,7 +206,7 @@ Thanks goes to these wonderful people ([emoji key](https://github.com/kentcdodds
...nfra-jakebolam "Infrastructure (Hosting, Build-Tools, etc)") | [<img src="https://avatars1.githubusercontent.com/u/39271619?v=4" width="100px;"/><br /><sub><b>Brandon Baksh</b></sub>](https://www.linkedin.com/in/brandonbaksh/)<br />[💻](https://github.com/tophat/yvm/commits?author=brandonbaksh "Code") | [<img src="https://avatars3.githubusercontent.com/u/2070398?v=4" width="100px;"/><br /><sub><b>Milan Milojic</b></sub>](https://github.com/nepodmitljivi)<br />[💻](https://github.com/tophat/yvm/commits?author=nepodmitljivi "Code") | [<img src="https://avatars2.githubusercontent.com/u/38886386?v=4" width="100px;"/><br /><sub><b>Umar Ahmed</b></sub>](https://github.com/umar-tophat)<br />[💻](https://github.com/tophat/yvm/commits?author=umar-tophat "Code") | [<img src="https://avatars0.githubusercontent.com/u/3258756?v=4" width="100px;"/><br /><sub><b>Nicholas Dujay</b></sub>](https://github.com/dat2)<br />[💻](https://github.com/tophat/yvm/commits?author=dat2 "Code") | [<img src="https://avatars0.githubusercontent.com/u/3996927?v=4" width="100px;"/><br /><sub><b>Aser Eldamaty</b></sub>](https://github.com/aeldamaty)<br />[💻](https://github.com/tophat/yvm/commits?author=aeldamaty "Code") |
```
jakebolam: there is an all contributors badge at the top which needs incrementing (it should have happened automatically)

src/yvm.js
```diff
@@ -35,20 +35,15 @@ argParser
...
-        if (maybeVersion) {
-            installVersion(maybeVersion).catch(error => {
-                log(error)
-                process.exit(1)
-            })
-            return
-        }
-        const [defaultVersion] = getSplitVersionAndArgs()
-        installVersion(defaultVersion).catch(error => {
+        const { fn: installFn, options } = command.latest
```
jakebolam: [nb] imo this is far harder to read and understand than the original code

README.md
iamogbz: I don't think it did. I just ran the command, will run it again, assuming it's idempotent

src/yvm.js
iamogbz: Refactored

src/util/log.js
```diff
@@ -15,4 +15,10 @@ log.info = function errorLog(...args) {
     }
 }
 
+log.colors = {
+    RED: '\x1b[31m%s\x1b[0m',
```
jakebolam: Thoughts on using chalk? https://github.com/tophat/yvm/pull/205/files

jakebolam:
LGTM?

PR #277, opened 2019-01-05
fix: invalid version number when running `yvm use` to auto-install version from `.yvmrc` file

umar-ahmed:
## Description

- On a fresh install of YVM, when I was running `yvm use` in a project that had a `.yvmrc` file, I expected it to auto-install the version and it failed instead
- After tracing the code path, I discovered that `installVersion` was being called with incorrect arguments

### DevQA Steps

- Run `make install`
- Create a folder with a `.yvmrc` file
- In the rc file, put a version number that you don't already have installed
- Run `yvm use`
- Verify that it downloads and installs the specified Yarn version and enables it

### DIFF
--- a/src/commands/install.js
+++ b/src/commands/install.js
@@ -167,7 +167,7 @@ const ensureVersionInstalled = (version, rootPath = yvmPath) => {
     if (fs.existsSync(yarnBinDir)) {
         return Promise.resolve()
     }
-    return installVersion(version, rootPath)
+    return installVersion({ version, rootPath })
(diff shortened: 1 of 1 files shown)


umar-ahmed:
@jakebolam it seems like Circle is using the latest yvm release which contains the bug that this PR is trying to fix. Anything that can be done about that from your end?

jakebolam:
@umar-ahmed thanks for catching this, I've pinned the version in CI temporarily.

PR #291, opened 2019-01-21
feat: yarn semantic version config

### LINKED ISSUE #276: Support .yarnversion file if .yvmrc is not present
Some other tools exist which use a .yarnversion file to indicate the version of yarn to use. YVM should be compatible with this existing practice.

When .yvmrc is not present, yvm should look for a .yarnversion file instead

### LINKED ISSUE #284: Support loosely specified versions (package.json engines version format)
It would be great to be able to specify the Yarn version more loosely, preferrably using the same format as in the engines property of package.json. (https://docs.npmjs.com/files/package.json#engines)

In my case, we use `yarn: 1.12` as our Yarn version in package.json engines. What I would like to do is to create a `.yvmrc.js` like this:

```
module.exports = require('./package.json').engines.yarn
```

in order to keep the yvm version in sync with the engine version. This works for specific versions, but gives me an error on `1.12` as yvm cannot infer the specific version to use.

It would maybe also be cool to have yvm use the engine version as fallback if there is no config file, as mentioned in #136, but I don't think this is the issue to discuss that. I just want to specify versions more loosely.

iamogbz:
## Description

- Closes #276
  - Adds `.yarnversion` to list of `cosmiconfig` search places
- Closes #284
  - Allows for semver style version declaration
  - Sets `engines.yarn` as `package.json` config property
- Addresses #205
  - Adds `chalk` dependency for showing colored log messages
- Addresses https://github.com/tophat/yvm/pull/275#discussion_r245470771
  - Replaces manual setting of console colors with chalk functions

### DevQA Prep

- Please `make install-watch`, this PR uses a new package

### DevQA Steps

- Modify the `.yvmrc` file to use semver style values
- `yvm use`
  - Should get the most recent matching locally installed version
  - Otherwise will install most recent version satisfying the config
- Rename the `.yvmrc` to `.yarnversion`
- `yvm use`
  - Should have no issue identifying the yarn version to use

## Comments
- Adds a bunch of tests


src/util/version.js
```diff
@@ -57,13 +70,26 @@ function setDefaultVersion({ version, yvmPath = defaultYvmPath }) {
 }
 
 function getRcFileVersion() {
-    const explorer = cosmiconfig('yvm')
+    const moduleName = 'yvm'
+    const explorer = cosmiconfig(moduleName, {
+        packageProp: 'engines.yarn',
```
iamogbz: Requires suggestion davidtheclark/cosmiconfig#164

src/util/version.js
```diff
@@ -57,13 +70,26 @@ function setDefaultVersion({ version, yvmPath = defaultYvmPath }) {
...
+        searchPlaces: [
+            'package.json',
+            `.${moduleName}rc`,
+            `.${moduleName}rc.json`,
+            `.${moduleName}rc.yaml`,
+            `.${moduleName}rc.yml`,
+            `.${moduleName}rc.js`,
+            `${moduleName}.config.js`,
```
iamogbz: Copied defaults from https://github.com/davidtheclark/cosmiconfig#searchplaces, can be reduced

src/commands/listRemote.js
```diff
@@ -22,7 +22,7 @@ const listRemoteCommand = () => {
                 localVersions,
             })
         })
-        .catch(error => log(error))
+        .catch(e => log.error(e))
```
jakebolam: [nb] `error`?

src/util/version.js
jakebolam: Is this blocked?

iamogbz: It might be, tried using a custom loader to get the `engines.yarn` prop but it seems like once it sees `'package.json'` it short circuits and doesn't use any specified custom loader

iamogbz: Might open another issue in the repo if I can't work around it. It's an easy enough implementation on their end but it might not be within the roadmap.

src/util/version.js
```diff
@@ -81,18 +107,17 @@ function getVersionInUse() {
 
 function getYarnVersions(yvmPath = defaultYvmPath) {
     const versionsPath = versionRootPath(yvmPath)
-    const prefixedVersionRegex = new RegExp(`^v${VERSION_REGEX.source}`)
     if (fs.existsSync(versionsPath)) {
         const files = fs.readdirSync(versionsPath)
         return files
-            .filter(name => prefixedVersionRegex.test(name))
+            .filter(name => name.startsWith('v') && isValidVersionString(name))
```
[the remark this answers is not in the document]
iamogbz: The previous regex was prefixed i.e. ``RegExp(`^v${VERSION_REGEX.source}`)``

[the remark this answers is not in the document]
iamogbz: The folder names follow the `v0.0.0` format

src/commands/listRemote.js
iamogbz:
```suggestion
        .catch(error => log.error(error))
```

src/util/version.js
iamogbz: https://github.com/davidtheclark/cosmiconfig/issues/164#issuecomment-456516124

PR #296, opened 2019-01-23
fix: print versions

iamogbz:
### DevQA Prep

### DIFF
--- a/src/util/version.js
+++ b/src/util/version.js
@@ -192,6 +192,7 @@ const printVersions = ({
 
         if (isCurrent) log.success(toLog)
         else if (isInstalled) log.notice(toLog)
+        else log(toLog)
 
         versionsMap[version] = toLog
     })
--- a/test/commands/list.test.js
+++ b/test/commands/list.test.js
@@ -1,19 +1,20 @@
 const fs = require('fs-extra')
 
-const list = require('../../src/commands/list')
+const version = require('../../src/util/version')
+const printVersions = jest.spyOn(version, 'printVersions')
 const {
     getExtractionPath,
     stripVersionPrefix,
     versionRootPath,
 } = require('../../src/util/utils')
-const { printVersions } = require('../../src/util/version')
+const list = require('../../src/commands/list')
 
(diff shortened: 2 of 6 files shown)


src/util/version.js
```diff
@@ -192,6 +192,7 @@ const printVersions = ({
 
         if (isCurrent) log.success(toLog)
         else if (isInstalled) log.notice(toLog)
+        else log(toLog)
```
iamogbz: Here be bug!

test/util/utils.test.js
```diff
@@ -1,39 +1,65 @@
...
-    it('Returns same number of versions passed to printVersion function', () => {
-        const versionsObject = printVersions({ list: versions, versionInUse })
+    afterEach(() => {
+        jest.resetAllMocks()
+    })
+
+    const verifyPrintedVersions = () => {
+        expect(log.mock.calls).toMatchSnapshot()
+        expect(log.notice.mock.calls).toMatchSnapshot()
+        expect(log.success.mock.calls).toMatchSnapshot()
```
iamogbz: Opted to use snapshots, rather than iterate over the calls

test/commands/list.test.js
```diff
@@ -43,19 +50,20 @@ describe('yvm list', () => {
...
 
-    it('Correctly highlights active yarn version', () => {
-        const listOutput = getList()
-        const versionsMap = printVersions({
-            list: listOutput,
-            message: 'test',
-            versionInUse: '1.7.0',
-        })
-        expect(versionsMap['1.7.0']).toContain(` \u2713 1.7.0`)
-        expect(versionsMap['1.6.0']).toContain(` - 1.6.0`)
```
iamogbz: This verification is now in `utils.test.js`, we just verify that `printVersions` is called with the right arguments.

iamogbz: Side note, it's a bit weird that `printVersion` tests were in the `utils.test.js` rather than `versions.test.js`

jakebolam:
Thanks for following up on this!

PR #299, opened 2019-01-26
feat: fish shell compatible

iamogbz:
## Description

- Move `yvm` shell function definition out of interactive block
- If running as script do nothing when no arguments passed

This allows sourcing of `yvm` using `bass`

### DevQA Prep

- Have `fish` shell
  - `$ brew install fish`
- Install [`edc/bass`](https://github.com/edc/bass)
- Please `make install-watch`, this PR uses a new package

### DevQA Steps

- In fish shell create wrapper function
```fish
function yvm
    bass . $YVM_DIR/yvm.sh ';' yvm $argv
end
```
- Run `yvm use 1.13.0`
- Should work as in bash shell

### Comments:

- Stop gap solution for fish shell users until #273 is merged. See https://github.com/tophat/yvm/issues/116#issuecomment-456819295.
- Similar technique recommended for using `nvm` in `fish` shell

### DIFF
--- a/src/yvm.sh
+++ b/src/yvm.sh
@@ -51,10 +51,11 @@ yvm_() {
     fi
 }
 
+yvm() {
+    yvm_ 'function' $@
+}
+
 if [ ${interactive} = 1 ]; then
-    yvm() {
-        yvm_ 'function' $@
-    }
     if ! type "node" > /dev/null; then
         yvm_err "YVM Could not find node executable."
         yvm_err "Please ensure your YVM env variables and sourcing are set below sourcing node/nvm in your .zshrc or .bashrc"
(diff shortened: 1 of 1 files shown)


eastenluis:
Does sourcing through bass update `$PATH` across terminals in this case?

iamogbz:
> Does sourcing through bass update `$PATH` across terminals in this case?

@eastenluis no, i don't think so. It just does it for the context of the command being run in the example `yvm` function

iamogbz:
@jakebolam can we get this in, currently have to revert my `yvm.sh` everytime I update self.

iamogbz:
Closing, feel free to reopen and merge.

PR #300, opened 2019-01-26
test: util log default

iamogbz:
## Description

- Allows spying on default log function
- Include console colors in log snapshots

## DevQA
- Verify happy path

### Comments:

- Only test changes


jakebolam:
Thanks for this @iamogbz !!

PR #290, opened 2019-01-19
feat: pipe sidio/stdout/stdin correctly between spawned yarn and parent

jakebolam:
## Description
Use native child process over shelljs exec
Fixes #286

### DevQA Steps
FAIL case:
- Install latest
- `yvm exec contributors:add`
- See input fail (and arrow keys)

PASS case:
- Check out this branch
- `make install`
- `yvm exec contributors:add`
- See input succeed (and arrow keys work!)

PR #302, opened 2019-01-29
feat: remove shelljs dependency

iamogbz:
## Description

- Removed shelljs dependency
- Updates remove logic to use `yarn --version` to check version in use
- Adds tests

### DevQA Prep

- Please `make install-watch`

### DevQA Steps

- `yvm i 1.8.0`
- `yvm ls` confirm version is listed
- `yvm rm 1.8.0` should remove version successfully
- `yvm ls` confirm version not listed
- `yvm use 1.13.0`
- `yvm ls` confirm version listed
- `yvm rm 1.13.0` should not remove version
- `yvm ls` confirm version still listed
- `yarn --version` confirm version still active

### Comments:

- Addresses comments from #290

### DIFF
--- a/src/util/version.js
+++ b/src/util/version.js
@@ -1,7 +1,7 @@
 const fs = require('fs')
 const path = require('path')
 const semver = require('semver')
-const { exec } = require('shelljs')
+const { execSync } = require('child_process')
 const cosmiconfig = require('cosmiconfig')
 
 const log = require('./log')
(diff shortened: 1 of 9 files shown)


src/util/version.js
```diff
@@ -92,17 +92,13 @@ function getRcFileVersion() {
...
-            { async: true, silent: true },
-            (code, stdout) => {
-                const versionInUse = stdout
-                resolve(versionInUse)
-            },
-        )
-    })
+async function getVersionInUse() {
+    try {
+        return String(execSync('yarn --version')).trim()
```
jakebolam: NB can use .to strong to chain the outputs

PR #318, opened 2019-03-10
fix: init yvm function in non interactive shell

### LINKED ISSUE #316: How to use yvm inside docker image?
I'm not sure that this behavior is a bug or expected behavior.

**What I'm trying to do?**
For development I'm creating a docker image and install everything needed into it. It includes needed languages like java/maven (sdkman), node/npm (nvm), but also vim, vscode, eclipse, android (build) tools. With vagrant It improves onboarding time for developers and easy way to try out new versions of used tools. Now I want to add yarn and struggle with `yvm`. For example, the minimal Dockerfile:

```
FROM ubuntu:18.04

RUN apt-get update && apt-get -yqq install curl unzip

SHELL ["bash", "--login", "-c"]

RUN curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.34.0/install.sh | PROFILE=~/.profile bash
RUN nvm install v10.15.3

RUN curl -fsSL https://raw.githubusercontent.com/tophat/yvm/v2.4.0/scripts/install.sh | sed 's/\.bashrc/\.profile/g' | INSTALL_VERSION="v2.4.0" bash
RUN yvm use 1.13.0
```
Because of non-interactive mode, I add initialization to `.profile` instead of `.bashrc`. But I get following error:

```
docker build --no-cache -t test .
[...]
Step 7/7 : RUN yvm use 1.13.0
 ---> Running in 928b75fe868c

  Usage: yvm [options] [command]

  Yarn Version Manager

  Options:

    -h, --help                        output usage information

  Commands:

    install|i [options] [version]     Install the specified version of Yarn.
[...]
    help [cmd]                        display help for [cmd]
bash: yvm: command not found
The command 'bash --login -c yvm use 1.13.0' returned a non-zero code: 127
```

You can add `-x` to `SHELL` to get more inside. It seems like yvm expects an argument in non-interactive mode even on initialization: `yvm_ 'script' $@`.

With interactive bash (`SHELL ["bash", "--login", "-i", "-c"]`) I can build image. But after creating a container there is no entry in PATH for yarn binary:
```
docker run --rm -it test bash --login
node --version
> v10.15.3
yvm version
> v2.4.0
yarn --version
> bash: yarn: command not found

env | grep PATH
> PATH=/root/.nvm/versions/node/v10.15.3/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

yvm use 1.13.0
> Now using yarn version 1.13.0
yarn --version
> 1.13.0
```

It would be nice that `yvm` doesn't require interactive shell like `nvm` and global yarn version an be set. Furthermore, `yvm` should set/switch yarn version, if it discovers `.yvmrc` in current or parent folders.

No text.

**Additional context**
No text.

iamogbz:
## Description

Should resolve #316
- Removes dependence on interactive mode
  - Refactor `yvm.sh` to function like `yvm.fish`

### DevQA Prep

- Please `make install-watch`

### DevQA Steps

- `yvm ls-remote` Should show list of available versions
- `bash --login -c 'yvm install <version>'` Any version from the list should work
- `yvm ls` Should show list of installed versions
- `bash --login -c 'yvm use <version>'` Any installed version should work
- `bash --login -c 'yvm exec --version'` Should show current yarn version

### DIFF
--- a/test/sanity-test.sh
+++ b/test/sanity-test.sh
@@ -1,7 +1,7 @@
-#!/bin/bash -e
+#!/bin/bash
 
 GREEN='\033[0;32m'
-RED='\031[0;32m'
+RED='\033[0;31m'
 NC='\033[0m' # No Color
 
 
@@ -23,11 +23,10 @@ fail() {
 }
 
 source ~/.yvm/yvm.sh
-YVM=~/.yvm/yvm.sh
 
 testing "yvm exec version command"
-${YVM} install 1.11.0
-test1_output=$(${YVM} exec 1.11.0 --version)
+yvm install 1.11.0
+test1_output=$(yvm exec 1.11.0 --version)
 if [[ ${test1_output} == "1.11.0" ]]; then
     pass
 else
@@ -35,9 +34,9 @@ else
 fi
 
 testing "yvm use"
-${YVM} install 1.7.0
-yvm_use # in place of yvm use
-test2_output=$(${YVM} exec --version)
+yvm install 1.7.0
+yvm use
+test2_output=$(yvm exec --version)
 if [[ ${test2_output} == "1.7.0" ]]; then
     pass
 else
--- a/src/yvm.fish
+++ b/src/yvm.fish
@@ -34,7 +34,7 @@ function yvm
         node "$YVM_DIR/yvm.js" $argv
     end
 
-    function yvm_init_fish
+    function yvm_init_sh
         if not type -q "node"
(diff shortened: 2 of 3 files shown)


test/sanity-test.sh
```diff
@@ -1,7 +1,7 @@
-#!/bin/bash -e
```
iamogbz: With this flag `source ~/.yvm/yvm.sh` causes sanity check to fail because `get-default-version` exits with 1 when none is set.

PR #323, opened 2019-03-11
feat: match nvm commands

### LINKED ISSUE #175: Add support for --latest, --stable/--lts, etc.
Something I love about `nvm` is that you can have it install `--latest` and `--lts` versions of `node`. This would be a great feature to see here, for convenience. `--stable` could be used instead of `--lts` in this case.

Currently I have to research what the latest or stable versions of `yarn` are.

### LINKED ISSUE #261: Provide a built-in fallback yarn version
Yvm does not work without a global version/.yvmrc file or version supplied on the command line

Provide a built-in fallback version. I.e. if it can't find a version it will use yarn 1.7.2.

**Additional context**
Came about when discussing #258

### LINKED ISSUE #262: Decide how much we'd like to reassemble nvm
Given that `yvm` is a lot like `nvm` in purpose etc, I think we should make a decision how much we want to mimic `nvm` from user's point of view. 

There's a clean advantage of being very much nvm-like, there will be much less steep learning curve, and less surprises. However going this way means that we won't be able to make a lot of UI decisions, as nvm is already quite stable and probably in most cases they've already made a decision we're currently making.

**Additional context**
This would influence all our user-facing interface - command names and semantics, options etc.

### LINKED ISSUE #328: update the website documentation to match the latest version of yvm
The website is the top hit on google.

Is the documentation on the website correct for the latest version of yvm?

https://yvm.js.org/docs/overview

iamogbz:
## Description

Should resolve #328, resolve #262, resolve #261 and resolve #175 
- [x] Remove implicit help command [`help [cmd]`](https://github.com/tj/commander.js/issues/683)
- [x] Add `uninstall` substitute for `remove` command
- [x] Replace `yvm version` with `yvm --version`
- [x] Add support for `nvm` like aliases
  - [x] Reserved aliases: latest, [stable](https://yarnpkg.com/lang/en/docs/install), system
  - [x] These can be used in place of version numbers in the commands
  - [x] Set `default` to `stable` if it's not set as fallback version
  - [x] Show if the alias resolved version can be installed or is local
- [x] Add `yvm current` which shows the current `yarn version`
  - [x] Indicates if it is a system version i.e. not installed by `yvm`, using the alias
  - [x] Update `which` command to show the `path` (BREAKING CHANGE)
- [x] Update documentation both `README` and `docs`.
- [x] Add tests

### DevQA Prep

- Please `make install`, this PR uses a new package

### DevQA Steps

**installation & sanity**
- `USE_LOCAL=true ./scripts/install.sh`
- `YVM_DIR=~/.yvm ./test/sanity-test.sh`

**commands regression**
- `yvm install`
- `yvm use`
- `yvm rm $(yarn --version)`
- `yvm exec --help`
- `yvm exec --version`
- `yvm exec contributors:add`

**which vs current**
- `yvm current` - should show the currently used yarn version
- `yvm which` - should show path to the version of yarn given

**alias command**
- `yvm alias`
- `yvm alias default 1.12.0` should replace default version
- `yvm alias default` & `yvm get-default-version` should match
- `yvm install latest` & `yvm install --latest` should match
- `yvm install stable` & `yvm install --stable` should match
- `yvm alias stable 1.7.0` should be rejected
- `yvm exec stable contributors:add`
- `yvm exec default contributors:add`
- `yvm uninstall stable` should remove the stable version installed
- `yvm which latest` - should show path to the latest version


jakebolam:
Awesome! Is this a breaking change/major version bump?

iamogbz:
@jakebolam yes it is, but only for the replacing `which` with `current`, might split that out of this PR when it's ready.

src/commands/current.js
```diff
@@ -7,7 +7,7 @@ const log = require('../util/log')
 const { versionRootPath } = require('../util/utils')
 const { yvmPath } = require('../util/path')
 
-const whichCommand = async path => {
+const currentCommand = async path => {
```
iamogbz: This is the only breaking change in this PR, might be split off

src/commands/exec.js
```diff
@@ -30,15 +29,16 @@ const runYarn = (version, extraArgs, rootPath) => {
...
+        await runYarn(...args)
+    } catch (e) {
+        log(e.message)
+        log.info(e.stack)
         process.exit(1)
-    })
+    }
+}
+
+module.exports = execCommand
```
iamogbz: Moves the `yvm-exec` command for consistency

test/sanity-test.sh
```diff
@@ -36,7 +36,7 @@ fi
 testing "yvm use"
 yvm install 1.7.0
 yvm use
-test2_output=$(yvm exec --version)
+test2_output=$(yvm exec -v)
```
iamogbz: This would be a breaking change, so needs to be resolved before this PR is ready.

webpack/webpack.config.base.js
```diff
@@ -9,7 +9,6 @@ module.exports = {
     mode: 'production',
     entry: {
         yvm: ['@babel/polyfill', './src/yvm.js'],
-        'yvm-exec': ['@babel/polyfill', './src/yvm-exec.js'],
```
iamogbz: Won't be needed anymore

src/util/alias.js
```diff
@@ -0,0 +1,49 @@
...
+const SYSTEM = 'system'
+const RESERVED = [LATEST, LTS, STABLE, SYSTEM]
+
+const resolveLatest = async () => {
+    const [version] = await getVersionsFromTags()
+    return version
+}
+
+const resolveStable = async () => {
+    // TODO: proper resolution of stable yarn version
```
iamogbz: Maybe grep the `https://yarnpkg.com/lang/en/docs/install/` page for the latest stable version? 😅

src/util/alias.js
```diff
@@ -0,0 +1,49 @@
...
+    return '1.13.0'
+}
+
+const resolveSystem = async () => {
+    const searchPaths = ['/usr/local/bin/yarn', '/usr/bin/yarn']
```
iamogbz: Very naive implementation! Maybe check for yarn versions in path as well?

src/yvm.js
```diff
@@ -97,18 +109,20 @@ argParser
         log.info('Checking for installed yarn versions...')
         const listVersions = require('./commands/list')
         listVersions()
-})
+    })
 
 argParser
-    .command('get-new-path [version]')
+    .command('get-new-path [version]', undefined, { noHelp: true })
```
iamogbz: Hides the command from the user help since it's internal

src/yvm.js
```diff
@@ -118,19 +132,52 @@ argParser
...
-    .action(version => {
-        const parsedVersionString = getValidVersionString(version)
-        if (!parsedVersionString) {
-            log('Error: Invalid version string supplied.')
+argParser.command('alias [<pattern>]', 'Show all aliases matching <pattern>')
+argParser.command(
+    'alias <name> <version>',
+    'Set an alias named <name> pointing to <version>',
+)
+argParser.command('alias', '', { noHelp: true }).action(async () => {
```
iamogbz: Done this way for overloading the `alias` command, it was tricky getting `commander` to allow this.

src/commands/exec.js
jakebolam: This file operates outside of the main file to allow commander to pass args through.

I.e. this solves
> Currently options shared between yarn and yvm cannot be passed to yarn in yvm exec. This is a limitation of commander the arg parse library.

src/commands/current.js
jakebolam: Are dropping commands breaking changes. E.g. remove?

src/yvm.js
```diff
@@ -43,40 +46,49 @@ argParser
...
-        process.exit(await remove(version))
-    })
+    .action(uninstallVersion)
 
-// commander has support for sub commands, and will load the file yvm-exec when this is run
-argParser.command(
-    'exec [version] [command]',
-    'Execute command using specified Yarn version.',
-)
+argParser
```
jakebolam: As per other comment, I think this code should stay in yvm-exec.js to allow commander to send args correctly to yarn

src/commands/current.js
iamogbz: Dropping commands are breaking changes, but `remove` isn't being removed `uninstall` was added and the help for `remove` indicates that `uninstall` is preferred

src/commands/exec.js
iamogbz: I'll double check, but that still didn't allow for `yvm --version` and `yvm exec --version` to exist simultaneously

src/yvm.js
iamogbz: So I double checked, once the '--version' option registered on `commander` it trumps calls to subcommands. Will try to find the issue for it, but yeah strange behaviour.

iamogbz: @jakebolam should be resolved now

src/commands/exec.js
iamogbz: `commander` treats `--version` special, but fixed it by only registering the version option when not an `exec` command

src/commands/current.js
iamogbz: Also changing the behaviour of `which` to show the path like it was initially intended I'll have to mark as breaking

src/util/utils.js
```diff
@@ -11,45 +13,51 @@ const getExtractionPath = (version, rootPath) =>
...
             }
+            return resolve(body)
         })
     })
-}
+})
+
+const getReleasesFromTags = memoize(async () => {
+    return getRequest(
+        'https://d236jo9e8rrdox.cloudfront.net/yarn-releases',
```
iamogbz: This seems to be cached somehow, even adding a `nonce` and `no-cache` header does not fetch the latest value

iamogbz: resolved in efc8cdc

jakebolam:
This is so awesome!

src/util/alias.js
```diff
@@ -0,0 +1,283 @@
...
+}
+
+const resolveAlias = memoize(
+    async ({
+        versionString,
+        currentAliases,
+        visitedAliases = [],
+        yvmPath = defaultYvmPath,
+    }) => {
+        const chain = visitedAliases.length ? visitedAliases.join(', ') : ''
```
mcataford: nb: `join(', ')` will already default to an empty string if `visitedAliases` has zero-length, this could just be 
```
const chain = visitedAliases.join(', ')
```
with the same result.

src/yvm.js
```diff
@@ -1,22 +1,19 @@
...
 
-const invalidCommandLog = () =>
-    log(
-        'You need to source yvm to use this command. run `source ~/.yvm/yvm.sh`',
-    )
+const messageNoYvm = 'Unable to determine yvm version'
+const messageSourceYvm =
+    'You need to source yvm to use this command. run `source ~/.yvm/yvm.sh`'
+const messageOptionalVersion = msg => `${msg}. Uses config file if available.`
+const signPosting = command => `See -> 'yvm ${command}'`
```
umar-tophat: (nb) Consider moving these to a constants file

src/yvm.js
```diff
@@ -31,52 +28,79 @@ if (!process.argv.includes('exec')) {
...
+        const { installVersion } = require('./commands/install')
+        try {
+            let version = maybeVersion
+            if (command.stable) {
+                version = alias.STABLE
+            } else if (command.latest) {
+                version = alias.LATEST
+            } else if (!version) {
+                version = (await getSplitVersionAndArgs())[0]
+            }
```
umar-tophat: (nb) move this to if-block to a helper function instead

src/util/alias.js
iamogbz:
```suggestion
        const chain = visitedAliases.join(', ')
```

src/yvm.js
```diff
@@ -31,52 +28,79 @@ if (!process.argv.includes('exec')) {
...
+argParser
+    .command('uninstall <version>')
     .alias('rm')
     .description('Uninstall the specified version of Yarn.')
-    .action(async version => {
-        log.info(`Removing yarn v${version}`)
-        const remove = require('./commands/remove')
-        process.exit(await remove(version))
-    })
+    .action(uninstallVersion)
```
umar-tophat: Can't you combine these two by adding multiple aliases?

```
argParser
	.command('uninstall <version>')
	.alias('rm')
	.alias('remove')
```

src/util/alias.js
```diff
@@ -0,0 +1,283 @@
...
+        if (dependants.length && !force && !recursive) {
+            log.notice(`The following aliases will be orphaned: ${dependants}
+Rerun with '--force' to remove just this alias
+Or with '--recursive' to remove all dependants`)
+        } else {
+            const namesToRemove = recursive
+                ? getAllDependants({ name, aliases })
+                : [name]
+            const newAliases = {}
+            for (const a of aliasNames) {
```
mcataford: `a` is a bit vague for a variable name, how about `aliasName` or `currentAlias`?

src/yvm.js
umar-tophat: Or you could just make the change directly, since this is going to a breaking change anyways?

README.md
```diff
@@ -66,13 +44,14 @@ curl -fsSL https://raw.githubusercontent.com/tophat/yvm/master/scripts/install.s
...
 Or to install a specific version:
+
 ```bash
 curl -fsSL https://raw.githubusercontent.com/tophat/yvm/master/scripts/install.sh | INSTALL_VERSION="v0.9.26" bash
 ```
 
 ### Manual installation
 
-Navigate to https://github.com/tophat/yvm/releases and download the `yvm.zip` file for the latest release to your home directory.
+Navigate to [github.com/tophat/yvm/releases](https://github.com/tophat/yvm/releases) and download the `yvm.zip` file for the latest release to your home directory.
```
aelkady: Can change to `Navigate to [releases](https://github.com/tophat/yvm/releases)`

test/commands/current.test.js
```diff
@@ -20,38 +20,47 @@ describe('yvm which command', () => {
...
         getVersionInUse.mockResolvedValueOnce('')
         const path = '/Users/tophat/.nvm/versions/node/v6.11.5/bin:'
-        expect(await which(path)).toBe(1)
+        expect(await current({ path })).toBe(1)
         expect(log).toHaveBeenCalledWith(
             expect.stringContaining(`yarn is NOT installed`),
         )
     })
     it('fails if environment path is not available', async () => {
         const oldPath = process.env.PATH
```
mcataford: nb: Ultimately, you could include this sort of thing in the `beforeEach` and `afterEach` (i.e. clone `process.env` before, restore it after.) That way, you ensure that nothing wonky can happen.

test/commands/current.test.js
```diff
@@ -20,38 +20,47 @@ describe('yvm which command', () => {
...
         delete process.env.PATH
-        expect(await which()).toBe(1)
+        expect(await current()).toBe(1)
         expect(log).toHaveBeenCalledWith(
             expect.stringContaining(`PATH not found`),
         )
         process.env.PATH = oldPath
     })
+    it('fails if fish environment path is not available', async () => {
+        const oldFishPath = process.env.FISH_USER_PATHS
```
mcataford: Same as the earlier comment on doing this in the before/after hooks.

test/commands/getNewPath.test.js
```diff
@@ -33,8 +37,60 @@ describe('getNewPath', () => {
...
+        ).toEqual(expectedPathString)
+    })
+
+    describe('fish', () => {
+        const FISH_ARRAY_DELIMINATOR = ' '
+        it('Returns a new fish_user_path string with a yarn directory prepended', () => {
+            const mockVersion = '1.7.0'
+            const mockRootPath = '/some/path'
+            const mockSplitPath = ['abc', 'def']
+            const mockPathString = mockSplitPath.join(FISH_ARRAY_DELIMINATOR)
```
mcataford: nb: s/DELIMINATOR/DELIMITER

test/commands/getNewPath.test.js
```diff
@@ -33,8 +37,60 @@ describe('getNewPath', () => {
...
+
+            const expectedPathString = [
+                `${mockRootPath}/versions/v${mockVersion}/bin`,
+                ...mockSplitPath,
+            ].join(FISH_ARRAY_DELIMINATOR)
```
mcataford: Same as ^

test/util/version.test.js
```diff
@@ -13,92 +15,134 @@ const {
...
-jest.mock('../../src/util/log')
-
 describe('yvm default version', () => {
     const mockYVMDir = '/mock-yvm-root-dir'
+    const defaultLog = jest.spyOn(log, 'default')
+    jest.spyOn(log, 'error')
+    jest.spyOn(log, 'info')
+    const resolveReserved = jest.spyOn(alias, 'resolveReserved')
     beforeEach(() => {
+        log()
```
mcataford: Mind elaborating on this?

src/yvm.js
```diff
@@ -31,52 +28,79 @@ if (!process.argv.includes('exec')) {
...
+    .action(async maybeVersion => {
+        log.info('Getting path to yarn version')
         const which = require('./commands/which')
-        process.exit(await which())
+        process.exit(await which({ version: maybeVersion }))
+    })
+
+argParser
+    .command('current')
+    .description('Display current active yarn version')
```
umar-tophat: (nb) `s/yarn/Yarn` for consistency

src/yvm.js
```diff
@@ -31,52 +28,79 @@ if (!process.argv.includes('exec')) {
...
-        if (command.latest) {
-            installLatest().catch(handleError)
-            return
-        }
-        const version = maybeVersion || await getSplitVersionAndArgs()[0]
-        installVersion({ version }).catch(handleError)
     })
 
+const uninstallVersion = async version => {
+    log.info(`Removing yarn v${version}`)
```
umar-tophat:
```suggestion
    log.info(`Removing Yarn v${version}`)
```

src/yvm.js
```diff
@@ -31,52 +28,79 @@ if (!process.argv.includes('exec')) {
...
-    .command('which')
-    .description(
-        'Display path to installed yarn version. Uses .yvmrc if available.',
-    )
-    .action(async () => {
-        log.info('Checking yarn version')
+    .command('which [version]')
+    .description(messageOptionalVersion`Display path to Yarn version`)
+    .action(async maybeVersion => {
+        log.info('Getting path to yarn version')
```
umar-tophat:
```suggestion
        log.info('Getting path to Yarn version')
```

src/yvm.js
```diff
@@ -31,52 +28,79 @@ if (!process.argv.includes('exec')) {
...
         const which = require('./commands/which')
-        process.exit(await which())
+        process.exit(await which({ version: maybeVersion }))
+    })
+
+argParser
+    .command('current')
+    .description('Display current active yarn version')
+    .action(async command => {
+        log.info('Checking yarn version')
```
umar-tophat:
```suggestion
        log.info('Checking Yarn version')
```

test/util/version.test.js
iamogbz:
```suggestion
        log() // see https://github.com/facebook/jest/issues/5792
```

src/yvm.js
iamogbz: Might as well

src/util/alias.js
```diff
@@ -0,0 +1,283 @@
...
+            try {
+                const pathToExec = path.join(pathEntry, 'yarn')
+                return String(
+                    execSync(`${pathToExec} --version 2> /dev/null`),
+                ).trim()
+            } catch (error) {
+                log.info(error.message)
+            }
+        }
+        return NOT_AVAILABLE
```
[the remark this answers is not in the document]
iamogbz: `UNRESOLVED` is a temporary placeholder to delay resolution until it is neccessary

`NOT_AVAILABLE` is a resolved value that indicated that the alias can not be installed

src/util/alias.js
```diff
@@ -0,0 +1,283 @@
...
+const SYSTEM = 'system'
+const UNRESOLVED = undefined
+const NOT_AVAILABLE = 'N/A'
+const DEFAULT_NAMES = [LATEST, STABLE, SYSTEM]
+const RESERVED_NAMES = [...DEFAULT_NAMES, LTS, NOT_AVAILABLE]
+
+const resolveLatest = memoize(async () => (await getVersionsFromTags())[0])
+
+const resolveStable = memoize(async () => {
+    const body = await getRequest('https://yarnpkg.com/lang/en/docs/install/')
```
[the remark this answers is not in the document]
iamogbz: No we don't, should that be included in this PR?

src/yvm.js
```diff
@@ -1,22 +1,19 @@
 const argParser = require('commander')
 
-const {
-    getSplitVersionAndArgs,
-    setDefaultVersion,
-    getDefaultVersion,
-    getValidVersionString,
-} = require('./util/version')
```
umar-tophat: What was the reasoning behind removing this from the global scope and moving it to the individual command handlers?

src/util/alias.js
```diff
@@ -0,0 +1,283 @@
...
+        (aliases, name) => Object.assign(aliases, { [name]: UNRESOLVED }),
+        {},
+    )
+}
+
+const getUserAliases = memoize(async (yvmPath = defaultYvmPath) => {
+    const aliases = { [DEFAULT]: STABLE }
+    const aliasStoragePath = path.join(yvmPath, STORAGE_FILE)
+    try {
+        const aliasesJSON = fs.readFileSync(aliasStoragePath, 'utf8')
```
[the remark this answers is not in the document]
iamogbz: Do not want to change the pattern of hidden files we have, but will test if you can require a json file without the json extension.

README.md
iamogbz:
```suggestion
Navigate to [yvm releases](https://github.com/tophat/yvm/releases) and download the `yvm.zip` file for the latest release to your home directory.
```

src/yvm.js
iamogbz: This allows the commands to run faster since the requirements are only done when the specific command needs them to run.

src/yvm.js
iamogbz: Leaving this in for now

src/yvm.js
```diff
@@ -97,40 +121,102 @@ argParser
...
     )
-    .action(version => {
-        const parsedVersionString = getValidVersionString(version)
-        if (!parsedVersionString) {
-            log('Error: Invalid version string supplied.')
+
+    const safeResolveVersion = async versionString =>
+        resolveVersion({ versionString }).catch(e => log.info(e.message))
+
+    if (typeof maybeVersion === 'string') {
```
umar-tophat: It seems like you are trying to do too many things by handling both versions of `alias` in one handler. I think it might makes sense to actually separate them even if there is a small bit of code duplication.

src/yvm.js
```diff
@@ -97,40 +121,102 @@ argParser
...
+        }
+        const noMatchFound = pattern.length > 0 && matchingAliases.length < 1
+        process.exit(noMatchFound ? 1 : 0)
+    }
+})
+
+argParser
+    .command('unalias <name>')
+    .option('-F, --force', 'Delete even if other aliases will be broken')
+    .option('-R, --recursive', 'Delete all dependant aliases as well')
```
umar-tophat: Why the switch to uppercase options for this command?

src/yvm.js
```diff
@@ -97,40 +121,102 @@ argParser
...
+    .description('Deletes the alias named <name>')
+    .action(async (name, { force, recursive }) => {
+        const alias = require('./util/alias')
+        const deleted = await alias.unsetAlias({ name, force, recursive })
+        if (deleted) {
+            log('Alias successfully deleted')
+        }
+        process.exit(deleted ? 0 : 1)
```
umar-tophat: Does it still exit with `1` if the alias you are trying to delete does not exist?

src/util/alias.js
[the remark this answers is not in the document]
brandonbaksh: Not a bad time to start :)

src/util/alias.js
brandonbaksh: can do for (const aliasName of aliasNameList)

src/yvm.js
mcataford: Seriousness, perhaps?

brandonbaksh: probably should support both. it is not uncommon to do lowercase imo

src/yvm.js
iamogbz: Yes

src/yvm.js
iamogbz: Uppercase was to make it more deliberate

src/util/alias.js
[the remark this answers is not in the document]
iamogbz: Done did!

src/yvm.js
mcataford: bitmoji

@brandonbaksh

Bessonov:
As I can see, nvm set first installed version as default. yvm set the newest version (not latest installed).

iamogbz:
@Bessonov yvm actually defaults to the latest stable version if you have no default version set. You should be able to overwrite it by setting your own default version.

Bessonov:
@iamogbz I've no problem with it, but this issue is about `match nvm commands` and this behavior is divergent.

PR #327, opened 2019-03-15
Preflight check on install dependencies

mcataford:
## Description
This PR adds pre-install checks to verify that all of the commands that the install process requires are installed and executable. In the case where it's not, it prints clear messaging about what package is missing and terminates the install process.

### DevQA Steps
You can temporarily remove `unzip` or `curl` from your PATH (or QA inside a Docker container that doesn't have those packages) -- trying to run the install script in those conditions should trigger the preflight check with the appropriate message.

If you run the install script in an environment where both `curl` and `unzip` are well-defined, you should be allowed through normally.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -16,7 +16,7 @@ INSTALL_DEPENDENCIES=( "unzip" "curl" )
 
 for package in "${INSTALL_DEPENDENCIES[@]}"
 do
-    if [ ! `eval "$package -v"` ]; then
+    if [ ! `eval "$package --version"` ]; then
         err_report_missing_dependency $package
         exit 1
     fi
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -12,7 +12,7 @@ err_report_missing_dependency() {
 
 trap 'err_report $LINENO' ERR
 
-INSTALL_DEPENDENCIES=( "unzips" "curl" )
+INSTALL_DEPENDENCIES=( "unzip" "curl" )
 
 for package in "${INSTALL_DEPENDENCIES[@]}"
 do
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -6,8 +6,22 @@ err_report() {
     echo "yvm failed to install, the command that failed was: https://github.com/tophat/yvm/blob/master/scripts/install.sh#L${1}"
 }
 
+err_report_missing_dependency() {
+    echo "The install cannot proceed due to a missing dependency. \"${1}\" is not installed."
```
iamogbz:
```suggestion
    echo "The install cannot proceed due to a missing dependency. \"${1}\" is not installed or available in your PATH."
```

scripts/install.sh
```diff
@@ -6,8 +6,22 @@ err_report() {
...
+}
+
 trap 'err_report $LINENO' ERR
 
+INSTALL_DEPENDENCIES=( "unzip" "curl" )
+
+for package in "${INSTALL_DEPENDENCIES[@]}"
+do
+    if [ ! `eval "$package -v"` ]; then
```
iamogbz: Have yet to QA, but `curl -v` does not exit with `0`

mcataford: That's a good point. `--version` however is common to both the packages specified, so it should return with a 0 if it's there.

```suggestion
    if [ ! `eval "$package --version"` ]; then
```

mcataford: I take it back, `--version` doesn't play well with `unzip`. I'll look into it tomorrow to find a better check.

mcataford: In a subsequent commit, I updated this to use `command -v <command>`, which should either give you the path to the executable or nothing, which works well with the `if` we have here.

mcataford:
@all-contributors please add @mcataford for code

PR #338, opened 2019-03-25
chore: deploy website

iamogbz:
### Checklist

### DevQA Prep

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,9 +93,10 @@ jobs:
       - save_env_cache
       - run: yarn build
       - persist_to_workspace:
-          root: ./
+          root: ../
           paths:
-            - build/
+            - docs/
+            - website/build/
   deploy_website:
     <<: *docker_defaults
     working_directory: ~/repo/website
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -93,9 +93,10 @@ jobs:
       - save_env_cache
       - run: yarn build
       - persist_to_workspace:
-          root: ./
+          root: ../
           paths:
-            - build/
+            - docs/
+            - website/build/
```
mcataford: `working_directory` already sets this stage up with `website/` as its root.

iamogbz: Changed the `persist_to_workspace` to be the parent folder. Testing that the `docusaurus_build` relies on the docs not the build folder.

mcataford: 👍

iamogbz: Probably don't need to persist this path then, also might as well just regenerate the docs

mcataford: Sounds good to me

Issue #348, opened 2019-04-01
Decouple node installer from local binaries


noahnu:
**Is your feature request related to a problem? Please describe.**

Currently yvm installer is strongly coupled to POSIX compliant systems (mkdir/rm) and common \*nix binaries (curl/unzip). yarn has cross-platform support, so by further decoupling yvm from specific platforms, we can move towards better cross-platform support as well.

**Describe the solution you'd like**

Replace all calls to binaries (mkdir/rm/curl) with built-in node equivalents.

**Describe alternatives you've considered**

- Compiling the standalone installer script via webpack so that we can import other packages (we would still manage total bundle size to keep it below some threshold). This is what yarn does.
- Support a number of "unzip" binaries, using the first binary available on the system.

iamogbz:
Thanks! @noahnu

PR #349, opened 2019-04-01
chore: remove curl, mkdir dependency

noahnu:
## Description

- Removes dependency on curl & mkdir in node installer, for #348.

### Checklist

### DIFF
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -1,10 +1,13 @@
 const os = require('os')
 const fs = require('fs')
 const { execSync } = require('child_process')
+const https = require('https')
+const path = require('path')
+const url = require('url')
 
 const log = (...args) => console.log(...args) // eslint-disable-line no-console
 
-const dependencies = ['curl', 'unzip']
+const dependencies = ['unzip']
 
 function preflightCheck(...dependencies) {
     const missing = []
@@ -67,9 +70,23 @@ function escapeRegExp(src) {
     return src.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
 }
 
-async function ensureDir(dirPath) {
-    if (fs.existsSync(dirPath)) return
-    execSync(`mkdir -p ${dirPath}`)
+/**
+ * Ensures a directory exists, creating the target
+ * directory and all ancestor directories if necessary.
+ *
+ * Equivalent to `mkdir -p $dirPath`.
+ *
+ * @param {string} dirPath
+ */
+function ensureDir(dirPath) {
+    if (!fs.existsSync(dirPath)) {
+        const directories = path.resolve(dirPath).split(path.sep)
+        let baseDir = directories[0] || '/'
+        for (const dirname of [...directories, '.']) {
+            if (!fs.existsSync(baseDir)) fs.mkdirSync(baseDir)
+            baseDir = path.join(baseDir, dirname)
+        }
+    }
 }
 
 function getTagAndUrlFromRelease(data) {
@@ -80,8 +97,8 @@ function getTagAndUrlFromRelease(data) {
     return { tagName, downloadUrl }
 }
 
-function getYvmVersion(versionTag, releasesApiUrl) {
-    const data = execSync(`curl -s ${releasesApiUrl}`)
+async function getYvmVersion(versionTag, releasesApiUrl) {
+    const data = await downloadFile({ source: releasesApiUrl })
     for (const release of JSON.parse(data)) {
         const { tagName } = getTagAndUrlFromRelease(release)
         if (tagName.match(new RegExp(versionTag))) {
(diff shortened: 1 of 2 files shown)


scripts/install.js
```diff
@@ -67,25 +71,58 @@ function escapeRegExp(src) {
...
 async function ensureDir(dirPath) {
-    if (fs.existsSync(dirPath)) return
-    execSync(`mkdir -p ${dirPath}`)
+    if (!fs.existsSync(dirPath)) {
+        const normalizedPath = path.normalize(path.relative('', dirPath))
+        ;[...normalizedPath.split(path.sep), '.'].reduce((basedir, dirname) => {
+            if (!fs.existsSync(basedir)) fs.mkdirSync(basedir)
+            return path.join(basedir, dirname)
+        })
+    }
```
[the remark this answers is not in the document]
noahnu: yeah I saw that..

scripts/install.js
```diff
@@ -67,25 +71,58 @@ function escapeRegExp(src) {
 }
 
 async function ensureDir(dirPath) {
-    if (fs.existsSync(dirPath)) return
-    execSync(`mkdir -p ${dirPath}`)
+    if (!fs.existsSync(dirPath)) {
+        const normalizedPath = path.normalize(path.relative('', dirPath))
```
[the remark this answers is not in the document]
noahnu: No, normalize.

`normalize` will remove redundant parts of the path so that the script iterates over less. It will also replace separation characters with the platform specific character.

`resolve` produces an absolute path. There's no guarantee we'll actually have permissions for each part of an absolute path. It's also potentially longer than a relative path (at least in our use case).

scripts/install.js
```diff
@@ -67,25 +71,58 @@ function escapeRegExp(src) {
...
+        ;[...normalizedPath.split(path.sep), '.'].reduce((basedir, dirname) => {
```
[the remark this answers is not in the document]
noahnu: In that case, it'd only reduce the number of directories checked by 1. We're not looping over all parts in the absolute path, but keeping it relative for this reason.

scripts/install.js
```diff
@@ -1,10 +1,14 @@
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
noahnu: huh.. I was actually getting issues in my local env with the missing hashbang. 

I'll revert that change in this PR and look into alternatives for solving the env issue I had.

😁 
```
curl -fsSL https://raw.githubusercontent.com/tophat/yvm/feat_348_decouple_node/scripts/install.js | tail +2 | node
```

scripts/install.js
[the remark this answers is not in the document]
iamogbz: Yeah it's only useful if you switch to `path.resolve` rather than `path.normalize`

scripts/install.js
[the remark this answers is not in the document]
iamogbz: Hmm, for example this
```js
path.normalize(path.relative('', '/usr/local/'))
```
would return 
```js
'../../../../usr/local'
```
so it would have to go through a bunch of unnecessary directories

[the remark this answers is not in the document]
iamogbz: About the permissions, we can't really control that, only inform the user why the install failed

iamogbz:
@noahnu any updates on this?

scripts/install.js
```diff
@@ -67,9 +70,23 @@ function escapeRegExp(src) {
...
+ * Ensures a directory exists, creating the target
+ * directory and all ancestor directories if necessary.
+ *
+ * Equivalent to `mkdir -p $dirPath`.
+ *
+ * @param {string} dirPath
+ */
+function ensureDir(dirPath) {
+    if (!fs.existsSync(dirPath)) {
+        const directories = path.resolve(dirPath).split(path.sep)
```
iamogbz: We can remove the current directory from the path

scripts/install.js
```diff
@@ -67,9 +70,23 @@ function escapeRegExp(src) {
...
+        let baseDir = directories[0] || '/'
+        for (const dirname of [...directories, '.']) {
```
iamogbz: When do we need the last `.`?

scripts/tests/install.test.js
```diff
@@ -105,6 +107,57 @@ describe('yvm install', () => {
...
+            } catch (err) {
+                expect(err.toString()).toMatch(/Only https scheme is supported/)
+            }
+        })
+
+        it('throws error on status error codes', async () => {
+            expect.assertions(1)
+            const mockGet = jest
+                .spyOn(https, 'get')
+                .mockImplementation((_, cb) => cb({ statusCode: 400 }))
```
iamogbz: nb. We can `spyOn` once and use `mockImplementationOnce` so we don't have to `mockRestore` at the end of every test, and just in the `afterAll`

scripts/install.js
noahnu: remnant of the old code, not necessary anymore

scripts/install.js
noahnu: I think the additional overhead from a few more path part checks is minimal. To remove the current directory, we'd have to be more intelligent with the dirpath resolution (and add quite a bit more logic I think to handle edge cases), unless you had something specific in mind?

scripts/tests/install.test.js
noahnu: I don't believe that would work.

If you moved the spyOn to a `beforeAll` and then did a mockRestore in an `afterAll`, we wouldn't be able to do the assertion on line 156 (assert called n times).

We could mockReset in an afterEach, however then we end up with more lines of code and I don't see the benefit of that approach.

I originally tried a mockRestore in an afterEach, however that messed up the other tests b/c there are jest mocks that are meant to persist until the end of the outer describe block. Those mock setups could be refactored so they're more isolated.

scripts/install.js
noahnu: The check is always done on the "base" directory. If you have a path like `/usr/local/bin` and split it by separator, you get:

```
const directories = ['', 'usr', 'local', 'bin']
```

The first element "" indicates a starting "/" and can therefore be reverted back to the root.

On the last iteration, dirname is "bin", however the basedir check will only ensure "/usr/local" exists. We add "." to force the basedir to contain the final directory.

iamogbz: We could move the `baseDir = path.join(baseDir, dirname)` line to before the exists check?

scripts/install.js
iamogbz: Yeah the overhead is minimal. We could
```js
const pwd = path.resolve()
const target = path.resolve(dirPath)
const directories = (target.startsWith(pwd) ? target.substr(pwd.length) : target).split(path.sep)
```
but that would only help sometimes

PR #357, opened 2019-04-06
fix: build new path without duplicates

### LINKED ISSUE #283: yvm doesn't set the path properly inside of tmux
**Describe the bug**
yvm does not set the PATH correctly inside tmux

**To Reproduce**
1. Enter a directory with a .yvmrc that is different from your current yarn version
1. `yvm use && yarn -v` should show the same version as your .yvmrc
1. Do the above inside a tmux session
1. yvm does not change the yarn version

Here is my PATH outside of tmux: 
```
/Users/michaelrose/.yvm/versions/v1.7.0/bin:/Users/michaelrose/.yarn/bin:/Users/michaelrose/.config/yarn/global/node_modules/.bin:/Users/michaelrose/.avn/bin:/usr/local/Cellar/pyenv-virtualenv/1.1.3/shims:/Users/michaelrose/.pyenv/shims:/Users/michaelrose/.nvm/versions/node/v8.11.3/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/sbin
```

And here is my path inside tmux:
```
/Users/michaelrose/.yarn/bin:/Users/michaelrose/.config/yarn/global/node_modules/.bin:/Users/michaelrose/.avn/bin:/usr/local/Cellar/pyenv-virtualenv/1.1.3/shims:/Users/michaelrose/.pyenv/shims:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/michaelrose/.yvm/versions/v1.7.0/bin:/Users/michaelrose/.yarn/bin:/Users/michaelrose/.config/yarn/global/node_modules/.bin:/Users/michaelrose/.avn/bin:/usr/local/Cellar/pyenv-virtualenv/1.1.3/shims:/Users/michaelrose/.pyenv/shims:/Users/michaelrose/.nvm/versions/node/v8.11.3/bin:/usr/local/sbin:/usr/local/sbin
```

It seem like in tmux, .yarn/bin gets added to the path before .yvm/versions. It's possible I could configure something locally on my machine to make this work as expected, but thought I'd file an issue anyway for tracking purposes and in case anyone comes across the same behaviour and is wondering what is going on.

**Expected behavior**

yvm should behave the same both inside and outside of tmux

**Desktop (please complete the following information):**
 - OS: MacOS 10.14.2
 - YVM Version: 1.0.9
 - Shell: bash
 - Node version: 8.11.3

iamogbz:
## Description

- Deduplicates path when building new path
- Always prepend yvm yarn version in path

### Checklist
- [x] This PR has sufficient testing

### DevQA Prep

- Please `make install`

### DevQA Steps

- Enter a directory with a .yvmrc that is different from your current yarn version
- yvm use && yarn -v should show the same version as your .yvmrc
- Do the above inside a tmux session
- yvm changes the yarn version

### Comments

- Fixes #283

### DIFF
--- a/src/commands/getNewPath.js
+++ b/src/commands/getNewPath.js
@@ -2,38 +2,18 @@ import path from 'path'
 
 import log from '../util/log'
 import { versionRootPath } from '../util/utils'
-import { getCurrentPath, getPathDelimiter, yvmPath } from '../util/path'
+import { getPathDelimiter, getPathEntries, yvmPath } from '../util/path'
 import { getSplitVersionAndArgs } from '../util/version'
 import { ensureVersionInstalled } from '../commands/install'
 
-export const buildNewPath = ({
-    version,
-    rootPath = yvmPath,
-    shell,
-    pathString,
-}) => {
+export const buildNewPath = ({ version, rootPath = yvmPath, shell }) => {
     const pathDelimiter = getPathDelimiter(shell)
-    const pathToUpdate = pathString || getCurrentPath(shell) || ''
-    const splitPath = pathToUpdate.split(pathDelimiter)
     const destPath = versionRootPath(rootPath)
-
     const newPathSegment = path.resolve(destPath, `v${version}`, 'bin')
-    let hadExistingPath = false
-
-    for (let i = 0; i < splitPath.length; i += 1) {
-        const pathSegment = splitPath[i]
-        if (pathSegment.startsWith(destPath)) {
-            splitPath[i] = newPathSegment
-            hadExistingPath = true
-            break
-        }
-    }
-
-    if (!hadExistingPath) {
-        splitPath.unshift(newPathSegment)
-    }
-
-    return splitPath.join(pathDelimiter).trim()
+    const isNotYvmYarnPath = pathSegment => !pathSegment.startsWith(destPath)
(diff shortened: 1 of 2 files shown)


src/commands/getNewPath.js
```diff
@@ -2,38 +2,18 @@ import path from 'path'
...
-    }
-
-    if (!hadExistingPath) {
-        splitPath.unshift(newPathSegment)
-    }
-
-    return splitPath.join(pathDelimiter).trim()
+    const isNotYvmYarnPath = pathSegment => !pathSegment.startsWith(destPath)
+    const nonYvmPathEntries = getPathEntries(shell).filter(isNotYvmYarnPath)
+    const updatedPath = [newPathSegment, ...new Set(nonYvmPathEntries)]
```
msrose: Does using a Set here potentially reorder other PATH entries? I'm not sure if order is guaranteed when iterating.

iamogbz: According to the [spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#Description) insertion order should be respected when accessing. But if there are duplicated only the first is preserved.

PR #368, opened 2019-04-14
feat: bundle yvm into a single js file

### LINKED ISSUE #348: Decouple node installer from local binaries

noahnu:
## Description

Closes #348

Combines all parts of yvm into a single downloadable `.js` release file that can be unpacked when installed.

### Checklist
- [x] This PR has updated documentation
- [x] This PR has sufficient testing

### DevQA Prep

- `rm -rf node_modules_production`
- `make clean`

### DevQA Steps

```sh
curl -s https://raw.githubusercontent.com/tophat/yvm/bundle_pkg/scripts/install.js | node
# should install the latest yvm release
```

```sh
curl -s https://raw.githubusercontent.com/tophat/yvm/bundle_pkg/scripts/install.js | INSTALL_VERSION='v2.4.3' node
# should install specified version of yvm
```

```sh
rm -rf ~/.yvm
# uninstall yvm (optional)
```

```sh
make install
# should install and configure shell using the current branch `bundle_pkg`
```

```sh
yvm --version
yvm ls-remote
# check yvm commands
```

```sh
which yarn
# should be pointing to yvm yarn shim
```

### Comments

- [x] Possibly split out into a couple PRs to minimize changes?
- [x] Fix installer to work with old tags (which expect zip)
- [x] Confirm no breaking changes
- [ ] Update brew tap release


scripts/install.js
```diff
@@ -241,26 +203,18 @@ async function run() {
         }
         await downloadFile({
             source: version.downloadUrl,
-            destination: paths.zip,
+            destination: path.join(paths.yvm, 'yvm.js'),
```
iamogbz: I think this should be kept backwards compatible

noahnu: Only reason we'd have to keep this backwards compatible is to deal with downloading an old version of yvm. Old yvm scripts should still pull the zip file (which we're publishing anyway). I think I do have to change some logic in the installer though to select the correct file to download.

iamogbz:
After #376 this PR scope should be more manageable

iamogbz:
@noahnu any update on this?

noahnu:
@iamogbz I'll update the PR in the next couple days, hopefully tonight.

.releaserc.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
         '@semantic-release/release-notes-generator',
         ["@semantic-release/github", {
             "assets": [
+                {"path": "artifacts/webpack_build/yvm.js", "name": "yvm.js"},
```
msrose: Based on talk in person, the first asset should be the zip file for compatibility with old installers

webpack/webpack.config.prod.js
```diff
@@ -1,11 +1,13 @@
 const path = require('path')
 const ZipFilesPlugin = require('webpack-zip-files-plugin')
+const LodashModuleReplacementPlugin = require('lodash-webpack-plugin')
 
 const { config, constants } = require('./webpack.config.base')
 
 module.exports = Object.assign(config, {
     plugins: [
         ...config.plugins,
+        new LodashModuleReplacementPlugin(),
```
msrose: I have run into nasty things in the past where all the lodash functions will be available in development, but then some won't get bundled into the production build because of how the plugin works. Just something to keep in mind.

package.json
```diff
@@ -35,8 +37,7 @@
     "cosmiconfig": "^5.1.0",
     "fs-extra": "^8.0.0",
     "kbpgp": "^2.1.0",
-    "lodash.escaperegexp": "^4.1.2",
-    "lodash.memoize": "^4.1.2",
+    "lodash": "^4.17.11",
```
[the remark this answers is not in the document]
iamogbz: Yeah but it's not expensive as evidenced by the reduced bundle size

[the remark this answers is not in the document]
noahnu: No. We're using the lodash webpack plugin so it's equivalent to importing the individual functions

webpack/webpack.config.prod.js
noahnu: I've used this plugin quite a bit. Do you recall any specific cases?

msrose: Basically you have to opt in to feature sets, it doesn't automatically include only the things you use, you have to make sure you've opted in to everything you're using https://github.com/lodash/lodash-webpack-plugin/blob/master/README.md#usage so if you used a collections function but didn't opt in to collections, it will work in dev but not in prod.

PR #400, opened 2019-05-29
feat: add undo yvm effects on the shell commands

iamogbz:
## Description

- Add nvm like `deactivate` and `unload` commands

### Checklist
- [x] This PR has updated documentation

### DevQA Prep

- Please `make install-watch`

### DevQA Steps

**`unload`**
- Open a new shell
- `echo $YVM_DIR` should show the location of yvm
- `which yarn` should point to the shimmed yvm yarn
- `yvm unload`
- `which yarn` should point to the system installed yarn or nothing if none
- `echo $YVM_DIR` should show nothing
- `yvm use` should fail

**`deactivate`**
- Open a new shell
- `echo $YVM_DIR` should show the location of yvm
- `which yarn` should point to the shimmed yvm yarn
- `yvm deactivate`
- `which yarn` should point to the system installed yarn or nothing if none
- `echo $YVM_DIR` should still show yvm location
- `yvm use`
- `which yarn` should point to the version of yarn in your `.yvmrc` file

### Comments

- See https://github.com/nvm-sh/nvm/issues/1399#issuecomment-279878984


src/util/path.js
```diff
@@ -1,6 +1,9 @@
 import os from 'os'
 import path from 'path'
 import fs from 'fs'
+import { negate } from 'lodash'
```
noahnu:
```suggestion
import negate from 'lodash/negate'
```

src/util/path.js
```diff
@@ -11,24 +14,39 @@ export const getPathDelimiter = shell => {
...
+export const isYvmPath = p => p && p.startsWith(yvmPath)
+const isYvmVersionPath = ({ p, rootPath = yvmPath }) =>
+    p.startsWith(versionRootPath(rootPath))
+const isShimPath = p => p && p.endsWith(shimRootPath(yvmPath))
+
 export const getPathEntries = shell =>
     getCurrentPath(shell).split(getPathDelimiter(shell))
 
+export const getNonYvmPathEntries = shell =>
+    getPathEntries(shell).filter(negate(isYvmPath))
```
noahnu: why use a function for negation?

iamogbz: I like the way it looks ☺️

PR #525, opened 2020-01-09
ci: add auto approve github action

iamogbz:
Adding auto approve action for dependabot spam


.github/workflows/auto-approve.yml
```diff
@@ -0,0 +1,12 @@
+
```
iamogbz:
```suggestion
```

Issue #771, opened 2020-09-17
yvm releases should be available whenever direct dependencies are updated


noahnu:
**Is your feature request related to a problem? Please describe.**

When direct dependencies of yvm are updated, for example a security vulnerability fix, currently we don't release a new version of yvm with the fixes in the yvm.js bundle.

**Describe the solution you'd like**

Update the release infrastructure to release on any direct dependency change. Might be able to tweak dependabot to use a different commit message prefix?

iamogbz:
I agree in the case of security vuln fixes but not the others. There's no need to cut a release if the user space has not effectively changed. Will look into the vuln case.

PR #772, opened 2020-09-17
chore: setup dependabot config

noahnu:
## Description

WIP.

Perhaps change "fix" to "build" and have that trigger a release? Not too familiar with how the auto release is setup in this project

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
```
iamogbz, line 1: Version 2 does not currently support auto merging to additional work will need to be done https://github.com/dependabot/dependabot-core/issues/1973

iamogbz: We can probably include the work needed for security vuln from #771 in here as well

PR #782, opened 2020-09-24
chore(docs): improve installation docs

jdbdnz:
## Description
Increase visibility of `--without-node` flag

### Comments
I've observed multiple people run the brew installation without reading the `--without-node` note first, then subsequently need to cancel the installation and start again. My hope is that by moving the note above the brew installation step people will read note _before_ running brew install rather than after.


valerierose:
Yes, I lost valuable time missing this note.

README.md
```diff
@@ -39,12 +39,12 @@ Node: >=10.0.0
 
 Installs the latest stable version.
 
+**NOTE**: Use the flag `--without-node` if you already have node installed.
+
 ```sh
 brew install tophat/bar/yvm
```
iamogbz, line 45: How about we just flip the condition

```suggestion
brew install tophat/bar/yvm --without-node
```

And in the note below, instruct users to remove the `--without-node` flag to get it to install node as well.

iamogbz (approved):

iamogbz:
@all-contributors add @jdbdnz for documentation

PR #889, opened 2020-12-17
chore: fixing codecov

akeshk:
Fixes codecov upload issue

## QA

- Ran the workflow on pull_request and it passed


.github/codecov.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# From https://github.com/decidim/decidim (AGPL-3.0)
+
+EVENT_PAYLOAD_FILE=$1
+
+PR_ID=`jq ".number // .check_run.pull_requests[0].number" $EVENT_PAYLOAD_FILE`
+GIT_SHA=`jq -r ".pull_request.head.sha // .check_run.head_sha // .after" $EVENT_PAYLOAD_FILE`
+
+if [ $PR_ID = "null" ]; then
+    bash <(curl -s https://codecov.io/bash) -C $GIT_SHA
```
noahnu, line 11: you can specify the artifact directory (whereever the junit files are written to) like this: https://github.com/tophat/monodeploy/blob/8167898783ed041bee2b2a536b732d6a1b5f2fc2/.github/codecov.sh#L11

.github/workflows/pull-request.yml
```diff
@@ -71,6 +71,10 @@ jobs:
         run: make node_modules
       - name: Tests
         run: make test-coverage
+      - name: Upload Coverage Report
```
noahnu, line 74: We're going to want to run codecov on push for master as well. Something similar to: https://github.com/tophat/monodeploy/blob/8167898783ed041bee2b2a536b732d6a1b5f2fc2/.github/workflows/master-branch.yml

.github/codecov.sh
akeshk: Done

.github/workflows/pull-request.yml
akeshk: Done

iamogbz: There should be a way not to duplicate the entire workflow. Maybe have this workflow run on both `push` and `pull_request` but some steps would be skipped based on the event.

noahnu: We can add more complex if conditionals to the steps to have it only run on push if the branch is master.

Otherwise GitHub actions really doesn't have a good mechanism to reuse workflows at the moment.

iamogbz: I think it's important because having separate workflows for basically the same thing can lead to drift and overhead

noahnu: I've seen it get quite messy with complex if conditionals. We might be able to simplify that though. Perhaps move it into an env var so we can do something like:

```
if: ${{ env.SHOULD_RUN }} == 'true'
```

or something like that..

Makefile
```diff
@@ -108,7 +108,6 @@ test-watch:
 .PHONY: test-coverage
 test-coverage:
 	$(JEST) --coverage
-	$(CODECOV)
```
iamogbz, line 111: I think you can remove the variable definition above as well

.github/codecov.sh
```diff
@@ -0,0 +1,14 @@
...
+EVENT_PAYLOAD_FILE=$1
+
+PR_ID=`jq ".number // .check_run.pull_requests[0].number" $EVENT_PAYLOAD_FILE`
+GIT_SHA=`jq -r ".pull_request.head.sha // .check_run.head_sha // .after" $EVENT_PAYLOAD_FILE`
+
+if [ $PR_ID = "null" ]; then
+    bash <(curl -s https://codecov.io/bash) -C $GIT_SHA -s $ARTIFACT_DIR
+else
+    bash <(curl -s https://codecov.io/bash) -C $GIT_SHA -P $PR_ID -s $ARTIFACT_DIR
+fi
```
iamogbz, line 14: No need for this file I think you can just use the [codecov github action](https://github.com/codecov/codecov-action)

